PR #2, opened 2020-06-12
Adds CI, requires CI tests to pass before merge to `master`

bobturneruk:


.travis.yml
```diff
@@ -0,0 +1,17 @@
...
+        - wget https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O miniconda.sh;
+        - bash miniconda.sh -b -p $HOME/miniconda
+        - source "$HOME/miniconda/etc/profile.d/conda.sh"
+        - hash -r
+        - conda config --set always_yes yes --set changeps1 no
+        - conda update -q conda
+        - conda create --name data_pipeline_api --file environment.yml
+        - conda activate data_pipeline_api
+      script:
+        - pytest ./data_pipeline_api/test_api.py
```
aflag: I think we can just add the upload conda script after this. It will build a noarch package. So that should just work fine. This is a bit different than it's usually done with travis, I think, but it should work for us. I think in most projects the release is triggered by pushing a version tag to github. However, I think integrating it like that is a little more involved.

The commit sha will still be inside the conda package, so that should be all we need for now in terms of being trackable. It'd be nice if we could auto-tag here somehow, but I think that'd require some git credentials.

aflag: Also, you'll probably need to install conda-build (`conda install conda-build`) before running the script.

bobturneruk: We'll need a separate issue to build via CI.

bobturneruk:
Not sure why this has done four calls to Travis - it has the same Webhooks as simple_network_sim. I've copied the approach on simple_network_sim as much as possible.

.travis.yml
bobturneruk: Issue to build via CI https://github.com/ScottishCovidResponse/SCRCIssueTracking/issues/532

mrow84 (approved):
I don't really know enough to be a useful reviewer, but I will approve anyway.

bobturneruk:
Cheers - what I intend this to do is:
- Build a container
- Clone the repo
- Build the conda environment in the container
- Run the tests
- Report coverage

I guess what you don't want are inappropriate checks that stop you merging stuff.

olexandr-konovalov (approved):
Looks good to me too. For infrastructural PRs, not changing functionality, completely fine to merge it rather sooner than later!

PR #4, opened 2020-06-16
Add `setup.py`

bobturneruk:


setup.py
```diff
@@ -0,0 +1,11 @@
+from setuptools import setup
+
+setup(
+   name='data_pipeline_api',
+   version='0.0.1',
+   description='Interacts with SCRC data pipeline',
+   author='SCRC',
+   author_email='scrc@glasgow.ac.uk',
+   packages=['data_pipeline_api'],
+#   install_requires=['h5py', 'toml'], # handle by conda, somehow?
```
aflag, line 10: I suggest we keep those as part of the setup.py and that we pin it to the specific version we want. After we have the wheel in place (and uploaded to pypi), we could create a conda package using `conda skeleton`. The only concern would be where to host it (it seems getting packages to conda-forge requires approval from the community, so it's potentially a slow process).

bobturneruk: What hosting infrastructure do we need?

aflag: We don't need anything ourselves. I believe we can just upload things to pypi.org and conda cloud. I have used pypi.org before, but not conda cloud. However, from my understanding, in conda cloud you can set up an organization and we could upload any conda packages we have there. Maybe we could also simplify things and have only the conda recipe, without uploading it to pypi.

bobturneruk:
@aflag - Will this PR be superseded by https://github.com/ScottishCovidResponse/data_pipeline_api/pull/5 ?

If so, I'll close it.

PR #6, opened 2020-06-17
Use git_url rather than path (save git info in conda package)

aflag:


conda/meta.yaml
```diff
@@ -3,7 +3,7 @@ package:
   version: 0.0.1
 
 source:
-  path: ..
+  git_url: ..
```
mrow84, line 6: Will `..` correctly locate the git repo, even in a local context? I presume it will identify the associated remote?

aflag: It will use the local git repo. That path is relative to where meta.yaml is. I could had set the actual git repo there pointing to master, but then the CI could package a different version from what it actually ran the tests for.

mrow84 (approved):

PR #7, opened 2020-06-23
Make tests run (on windows)

bobturneruk:


bobturneruk:
For illustrative purposes only. For now.

bobturneruk:
Had a brief chat with @aflag - think this folder structure will play nicer with pytest on more platforms.

mrow84:
I don't have any useful opinions of my own about this, but I thought it worth noting that I changed to the `src/` layout specifically on the recommendations of [this section](https://docs.pytest.org/en/latest/goodpractices.html#tests-outside-application-code) of the pytest documentation.

bobturneruk:
I think this is less relevant given that we're using conda environments. However, if we switch to PyPi this may change. I've asked @aflag for a review to get his opinion.

bobturneruk:
Adding this `conftest.py` to `tests/` would also work but slightly undermine the point of having `src` there in the first place.
```
import sys

sys.path.append("src")
```

ZedThree:
When using the `src/`, `tests/` directory layout, one would do a local install of the package and test that, so you shouldn't need to change `sys.path`. When using developing locally, all you'd need to do is `pip install -e .`. On CI platforms, you'd install the package, some of them do this automatically for you.

Also, probably not from this PR, but I get one test failure (`example_data/human/demographics/simple_network_sim` is missing), and one warning from h5py (missing mode from file open):

```
========================================= test session starts ==========================================
platform linux -- Python 3.8.3, pytest-5.4.3, py-1.8.2, pluggy-0.13.1
rootdir: /home/peter/Codes/Covid/data_pipeline_api
collected 12 items                                                                                     

tests/test_api.py ...F.                                                                          [ 41%]
tests/test_file_system_data_access.py ...                                                        [ 66%]
tests/test_parameter_file.py ....                                                                [100%]

=============================================== FAILURES ===============================================
_______________________________________ test_session_read_table ________________________________________

    def test_session_read_table():
        api = API(FileSystemDataAccess(DATA_PATH, None))
>       table = api.read_table("human/demographics/simple_network_sim")

tests/test_api.py:36: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
data_pipeline_api/api.py:164: in read_table
    self._read_parameter_version_bytes(
data_pipeline_api/api.py:91: in _read_parameter_version_bytes
    parameter_version = self._data_access.get_parameter_version(parameter, version)
data_pipeline_api/file_system_data_access.py:71: in get_parameter_version
    for p in parameter_path.iterdir():
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = PosixPath('/home/peter/Codes/Covid/data_pipeline_api/example_data/human/demographics/simple_network_sim')

    def iterdir(self):
        """Iterate over the files in this directory.  Does not yield any
        result for the special paths '.' and '..'.
        """
        if self._closed:
            self._raise_closed()
>       for name in self._accessor.listdir(self):
E       FileNotFoundError: [Errno 2] No such file or directory: '/home/peter/Codes/Covid/data_pipeline_api/example_data/human/demographics/simple_network_sim'

/usr/lib64/python3.8/pathlib.py:1118: FileNotFoundError
=========================================== warnings summary ===========================================
tests/test_api.py::test_session_read_matrix
tests/test_api.py::test_log_reads
  /home/peter/Codes/Covid/data_pipeline_api/data_pipeline_api/api.py:145: H5pyDeprecationWarning: The default file mode will change to 'r' (read-only) in h5py 3.0. To suppress this warning, pass the mode you need to h5py.File(), or set the global default h5.get_config().default_file_mode, or set the environment variable H5PY_DEFAULT_READONLY=1. Available modes are: 'r', 'r+', 'w', 'w-'/'x', 'a'. See the docs for details.
    dataset_root: h5py.File = h5py.File(

-- Docs: https://docs.pytest.org/en/latest/warnings.html
======================================= short test summary info ========================================
FAILED tests/test_api.py::test_session_read_table - FileNotFoundError: [Errno 2] No such file or dire...
=============================== 1 failed, 11 passed, 2 warnings in 0.90s ===============================
```

mrow84:
Regarding the test failures, #8 has (hopefully) more stable tests, so it might be worth evaluating these changes in the context of that branch.

bobturneruk:
@ZedThree - cheers - I think the issue is that we're either not doing, or we've moved away from pip install in favour of something conda-based. I think this means we're in the process of dropping `setup.py`. 

I find all the tests pass on https://github.com/ScottishCovidResponse/data_pipeline_api/pull/8 if I get `src` on path.

aflag (approved):

aflag:
`setup.py` is still needed, that's what the conda recipe uses. We currently have a bit of duplication of dependencies, but I think that could go away if we start creating a wheel and pushing it to PyPI. BTW, I didn't realise you were having problems with setup.py. Indeed, we are missing a line in there: ```package_dir = {'': 'src'},```

aflag:
I've added {'': 'src'} on master, but feel free to remove it. I particularly like the idea of not using the src directory.

bobturneruk:
> I don't have any useful opinions of my own about this, but I thought it worth noting that I changed to the `src/` layout specifically on the recommendations of [this section](https://docs.pytest.org/en/latest/goodpractices.html#tests-outside-application-code) of the pytest documentation.

@aflag - do you have a stance on this? We're moving towards making Python packages. My reading is that the `src` structure is intended to help ensure that the installed version of a package is tested, not the code in the folder in the repo. On simple network sim we don't have a package, so we want to test the code in the folder. Maybe things are different, here? Maybe it depends on how we call the tests in the CI?

aflag:
Hm, I see the reasoning. I've seen people actually deleting the code folder after the wheel or egg is installed. I'm not 100% that having the code inside `src` will solve that problem, because the testing tool should be able to run locally too, so you have to do something explicitly to make it ignore the src folder when running in the CI. But I get how it would be easier to guarantee that in such case.

Anyway, given we don't ship any files inside the wheel or C dependencies, I think the difference between testing the installed code rather than the code in the `src` directory is rather small.

bobturneruk:
OK. Let's leave this unmerged for a bit in case there are any concrete objections to the changes. Otherwise it's helpful to end the discussion so it doesn't hamper https://github.com/ScottishCovidResponse/SCRCIssueTracking/issues/504 and https://github.com/ScottishCovidResponse/SCRCIssueTracking/issues/419. Please speak up if you think we should keep `src`!

mrow84:
I'm happy to bin `src`. Unfortunately master has moved on quite a bit since you opened this, so I think you'll have to redo the changes, sorry 😩 .

bobturneruk:
Yeah - just tried to rebase and my laptop exploded. No worries!

bobturneruk:
Superseded by https://github.com/ScottishCovidResponse/data_pipeline_api/pull/13. This branch will not be merged.

PR #8, opened 2020-06-24
Scrc 467 split api

mrow84:
The aim of this change was to split the "lower level" file API, which implements a generic metadata-based file handling and tracking scheme, from the "higher level" APIs which deal with reading and writing particular file formats.

The main job after merging this will be to flesh out the "standard API". This should be relatively straight-forward, as I had crude implementations of most of that working in a previous version. I will at that time add examples of all the standard file formats.

- [x] squash before merging


environment.yml
```diff
@@ -10,6 +10,8 @@ dependencies:
   - toml=0.9.4
   - h5py=2.9.0
   - scipy=1.4.1
+  - yaml=5.3.1
```
bobturneruk: Should this be `pyyaml=5.3.1`? I can't see `yaml` on conda at 5.3.1 whereas 5.3.1 is the latest `pyyaml`.

mrow84: It certainly should, will fix

src/data_pipeline_api/file_api.py
```diff
@@ -0,0 +1,178 @@
...
+
+        # Compute a run_id.
+        with open(config_filename, "rb") as file:
+            m = sha1(file.read())
+        m.update(str(datetime.now()).encode())
+        self.run_id = m.hexdigest()
+
+        try:
+            with open(config_filename) as config_file:
+                config = yaml.load(config_file)
```
ZedThree: I get a warning:

```
YAMLLoadWarning: calling yaml.load() without Loader=... is deprecated, as the default Loader is unsafe. Please read https://msg.pyyaml.org/load for full details.
```

examples/data_access.py
```diff
@@ -1,13 +1,24 @@
...
+[ ] Add up-front verification, to make sure we are not about to start a run that we
+    cannot complete.
+[ ] We shouldn't fail on overwriting a file, but rather a component?
+"""
 
-api.write_table(population, "output/table", version=1)
-api.close()
+with SimpleNetworkSimAPI(
+    "repos/data_pipeline_api/examples/test_data_2/config.yaml",
+) as api:
```
ZedThree, line 21: Nice! Definitely think a context manager is more ergonomic over `api.close()`.

For the C++ bindings, we should be able to use RAII to achieve the same thing

src/data_pipeline_api/file_api.py
```diff
@@ -0,0 +1,178 @@
...
+        if path.exists():
+            file = open(path, mode="r+b")
+        else:
+            if not path.parent.exists():
+                path.parent.mkdir(parents=True)
+            file = open(path, mode="w+b")
+        # Wrap the file close method with something to record the file access.
+        close_file = file.close
+
+        @wraps(close_file)
```
aflag, line 147: I'm curious, wouldn't `"wraps(file.close)` work?

src/data_pipeline_api/file_api.py
```diff
@@ -0,0 +1,178 @@
...
+                    write_metadata[MetadataKey.extension],
+                )
+            )
+
+        path = self.normalised_data_directory / write_metadata[MetadataKey.filename]
+        if path.exists():
+            file = open(path, mode="r+b")
+        else:
+            if not path.parent.exists():
+                path.parent.mkdir(parents=True)
```
ZedThree:
```suggestion
            path.parent.mkdir(parents=True, exist_ok=True)
```

ZedThree (approved):
LGTM. yaml.load warning should probably be fixed though

src/data_pipeline_api/file_api.py
mrow84: fixed

src/data_pipeline_api/simple_network_sim_api.py
```diff
@@ -0,0 +1,13 @@
+from io import TextIOWrapper
+from pathlib import Path
+from pandas import read_csv, DataFrame
+from data_pipeline_api.file_api import FileAPI
+
+class SimpleNetworkSimAPI(FileAPI):
```
aflag, line 6: Using composition rather than inheritance here would allow us to hide away the the lower level methods no one should be using directly. What do you think?

src/data_pipeline_api/metadata.py
```diff
@@ -0,0 +1,22 @@
+from typing import Mapping, Optional
+from pathlib import Path
+from semver import VersionInfo
```
aflag: This seems unused

environment.yml
```diff
@@ -10,6 +10,8 @@ dependencies:
   - toml=0.9.4
   - h5py=2.9.0
   - scipy=1.4.1
+  - pyyaml=5.3.1
+  - semver=2.9.0
```
rtblackwell, line 14: any reason for this version rather than the latest?

src/data_pipeline_api/file_api.py
```diff
@@ -0,0 +1,177 @@
...
+
+        # Compute a run_id.
+        with open(config_filename, "rb") as file:
+            m = sha1(file.read())
+        m.update(str(datetime.now()).encode())
+        self.run_id = m.hexdigest()
+
+        try:
+            with open(config_filename) as config_file:
+                config = yaml.safe_load(config_file)
```
rtblackwell, line 30: just to check my understanding - the safe loading is OK because we're assuming everything is going to be dicts/lists/strs and nothing more complex?

src/data_pipeline_api/file_api.py
```diff
@@ -0,0 +1,177 @@
...
+            self.access_log_path = config_filename.parent / self.access_log_path
+        if self.access_log_path.exists():
+            raise ValueError(f"{self.access_log_path} already exists")
+
+        with open(
+            self.normalised_data_directory / "metadata.yaml"
+        ) as metadata_store_file:
+            self._metadata_store = MetadataStore(yaml.safe_load(metadata_store_file))
+
+        self._open_timestamp = datetime.now()
```
rtblackwell: call datetime.now() once and use it for the sha and the open timestamp? I wonder if it could be useful to know the datetime that went into the sha for any debugging etc

environment.yml
mrow84: I wrote this in manually, and just put in the version that I got from a conda install. Is there a newer conda version? I have no idea whether the version schemes are all consistent.

src/data_pipeline_api/file_api.py
mrow84: That is my understanding as well 🤷

environment.yml
bobturneruk: I think this explicitly specifies the latest versions current on conda, if that helps?

mrow84: Apparently there is a newer version on a different conda channel, but I've no idea whether using a different channel is a good idea. @bobturneruk, @aflag ?

bobturneruk: I think it depends - do we want features / bug fixes from the newer version? Is it on a channel we trust?

rtblackwell (approved):

environment.yml
aflag: I think having something like pyyaml>=5.3 for a library should be fine. In the actual application that uses the library we'd want to have some sort of pinning (the spec-file.txt net work sim's case, for instace.

aflag: BTW, can you rebase master onto your branch, it'd be nice to also update the meta.yaml file

aflag (approved):

src/data_pipeline_api/simple_network_sim_api.py
mrow84: Yes I greatly prefer composition, but I left it like this precisely so that people could access those methods in this uncertain state - once we have pinned things down I think we should do what you suggest.

src/data_pipeline_api/metadata.py
mrow84: fixed

src/data_pipeline_api/file_api.py
mrow84: yes it would, though we need `close_file` for the invocation inside our redefined `close`

PR #9, opened 2020-06-24
Add CMake support to C++ bindings

ZedThree:
This adds a pretty basic CMake file for the C++ bindings branch. The pybind11 submodule basically takes care of all the fun of finding python for us.

Unfortunately pybind11 does a couple of things strangely, so it doesn't currently seem possible to use the build directory directly. This means the pipeline bindings need to be installed:

    cmake . -B build -DCMAKE_INSTALL_PREFIX=$(pwd)/build/install

A minimal `CMakeLists.txt` to use this would be:

```cmake
cmake_minimum_required(VERSION 3.12)

project(data_pipeline_cpp_test
  VERSION "0.1.0"
  LANGUAGES CXX)

find_package(data_pipeline_cpp REQUIRED)

add_executable(test_datapipeline
  test_datapipeline.cc)
target_link_libraries(test_datapipeline data_pipeline_cpp::data_pipeline_cpp)
```

and configured like:

    cmake . -B build -DCMAKE_PREFIX_PATH=/path/to/installed/bindings

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "bindings/cpp/pybind11"]
+	path = bindings/cpp/pybind11
+	url = https://github.com/pybind/pybind11.git
--- a/bindings/cpp/CMakeLists.txt
+++ b/bindings/cpp/CMakeLists.txt
@@ -1,15 +1,61 @@
 cmake_minimum_required(VERSION 3.12)
 
 project(data_pipeline_cpp
+  VERSION "0.1.0"
   LANGUAGES CXX)
 
+set(PYBIND11_INSTALL ON CACHE BOOL "")
 add_subdirectory(pybind11)
+
+set(DATA_PIPELINE_HEADERS table.hh datapipeline.hh)
 
 add_library(data_pipeline_cpp
   table.cc datapipeline.cc
-  table.hh datapipeline.hh)
+  ${DATA_PIPELINE_HEADERS})
 target_link_libraries(data_pipeline_cpp PUBLIC pybind11::embed)
 
 add_executable(test_datapipeline
   test_datapipeline.cc)
 target_link_libraries(test_datapipeline data_pipeline_cpp)
+
+##################################################
+# Installation
+
+include(GNUInstallDirs)
+install(TARGETS data_pipeline_cpp
+  EXPORT data_pipeline_cppTargets
+  LIBRARY DESTINATION "${CMAKE_INSTALL_LIBDIR}"
+  ARCHIVE DESTINATION "${CMAKE_INSTALL_LIBDIR}"
+  RUNTIME DESTINATION "${CMAKE_INSTALL_BINDIR}"
+  INCLUDES DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}"
+  )
+install(FILES ${DATA_PIPELINE_HEADERS} DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}")
+
+include(CMakePackageConfigHelpers)
+write_basic_package_version_file(
+  data_pipeline_cppConfigVersion.cmake
+  VERSION ${PACKAGE_VERSION}
+  COMPATIBILITY SameMajorVersion
+  )
+
+# This is a workaround for pybind11 not currently installing its targets
+# Definitely should be removed if that changes
+install(EXPORT "pybind11Targets"
+  NAMESPACE "pybind11::"
+  DESTINATION ${PYBIND11_CMAKECONFIG_INSTALL_DIR})
+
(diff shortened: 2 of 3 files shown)


bindings/cpp/CMakeLists.txt
```diff
@@ -1,15 +1,61 @@
...
 target_link_libraries(data_pipeline_cpp PUBLIC pybind11::embed)
 
 add_executable(test_datapipeline
   test_datapipeline.cc)
 target_link_libraries(test_datapipeline data_pipeline_cpp)
+
+##################################################
+# Installation
+
+include(GNUInstallDirs)
```
qingfengxia, line 24: I had no experience/knowledge on `include(GNUInstallDirs)`,  usually i used CPack_*  to deal with installation

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "bindings/cpp/pybind11"]
+	path = bindings/cpp/pybind11
```
qingfengxia, line 2: no ideal if another submodule may be introduced,  it may make sense to  put all  third-party lib   in a `third-party`folder, anyway, this can be done in the future.

qingfengxia (commented):
it is a good feature,  it ease the process of compiling pybind11::embed

qingfengxia:
I will have a look on the issue 
Unfortunately pybind11 does a couple of things strangely, so it doesn't currently seem possible to use the build directory directly. This means the pipeline bindings need to be installed:

cmake . -B build -DCMAKE_INSTALL_PREFIX=$(pwd)/build/install

I have not run into any problem ;  this may do the trick
```

## put all targets in bin and lib
set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}/bin)
set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}/lib)
link_directories(${CMAKE_LIBRARY_OUTPUT_DIRECTORY})
```

qingfengxia:
here is my test cmake, on Ubuntu 20.04
```
#############################
cmake_minimum_required(VERSION 3.12)

project(data_pipeline_cpp
  VERSION "0.1.0"
  LANGUAGES CXX)

###############################################

# clone only the branch, wihch conflicts with master 
# git clone -b cppbindings --single-branch git@github.com:ScottishCovidResponse/data_pipeline_api.git
# install all python modules , see Ian's Readme.md
# using system installed pybind11 ubuntu 20.04 and my mini CMakeLists.txt
# cmake .. -DPYTHON_EXECUTABLE:FILEPATH=$(which python3)
# export PYTHONPATH=$PWD/../../src:$PYTHONPATH
# run the test_datapipeline 

## put all targets in bin and lib
set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}/bin)
set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}/lib)
link_directories(${CMAKE_LIBRARY_OUTPUT_DIRECTORY})

find_package(pybind11 QUIET)  # systemwide installation by package manager  `/usr/include/`
if(NOT pybind11_FOUND)
    message("===============================================================\n"
            "pybind11 not found on system, use the downloaded version \n"
            "===============================================================\n")
    add_subdirectory(pybind11)  # there is no need to find_package
    set(pybind11_FOUND ON)
endif()

#set(PYBIND11_INSTALL ON CACHE BOOL "")
#add_subdirectory(pybind11)

set(DATA_PIPELINE_HEADERS table.hh datapipeline.hh)

add_library(data_pipeline_cpp
  table.cc datapipeline.cc
  ${DATA_PIPELINE_HEADERS})
target_link_libraries(data_pipeline_cpp PUBLIC pybind11::embed)

add_executable(test_datapipeline
  test_datapipeline.cc)
target_link_libraries(test_datapipeline data_pipeline_cpp)

```

ianhinder:
Hi! Sorry to be late with this.  I'm not a cmake guru; the bindings makefile is quite simple, and assumes that you're going to use the python called "python3" on your PATH.  This should be configurable in future (i.e. it should use $(PYTHON)).  It then gets all the flags it needs from python3-config.  My worry with this PR is only that the submodule used to get the pybind header might be a different version of pybind than the one used by the bindings. Could this happen?  I think it is cleaner to use the available pybind installation whatever is being used, pip, conda, system, etc.  That way everything is consistent. Or it's possible I've misunderstood and everything is fine!

qingfengxia:
cppbindings-v2 should have the cmake merged soon today by Peter Fox,  partial come from this PR,  so this PR may be outdated,  some installation related code may be worth of another PR in later stage.

PR #14, opened 2020-06-26
Scrc 467 update standard api

mrow84:
This adds something approximating the standard API spec. I may do some more clean up before merging, but this should work, and can probably be usefully integrated against.

- [x] squash before merge


bobturneruk:
Is the aspiration here that this PR is specification-compliant before merge? If so, I suggest that the specification (once agreed) is included in this PR such that it can change in lockstep with the implementation.

mrow84:
> Is the aspiration here that this PR is specification-compliant before merge? If so, I suggest that the specification (once agreed) is included in this PR such that it can change in lockstep with the implementation.

Whilst that wasn't a level of rigour that I was aspiring to, it does seem like a good idea - I presume you mean for me to reference the file and maybe give it a version number?

bobturneruk:
I'd suggest turning into markdown and adding it to this repo.

mrow84:
> I'd suggest turning into markdown and adding it to this repo.

Hmm ok - I guess we can do that in snapshots, but I wouldn't say the spec is necessarily "done", and the reason for doing it on teams was that it was easier to comment there than on markdown pages (in the wiki, at least).

bobturneruk:
Yeah - I agree it's easier to comment on teams. I think it should only go in here if it's pretty stable.

mrow84:
> Yeah - I agree it's easier to comment on teams. I think it should only go in here if it's pretty stable.

Ok, well I intend on doing a "finalisation" PR once we've pinned down the spec, so I'll include it at that point.

data_pipeline_api/file_api.py
```diff
@@ -19,15 +19,17 @@ def __init__(
...
-        with open(config_filename, "rb") as file:
-            m = sha1(file.read())
-        m.update(str(self._open_timestamp).encode())
-        self.run_id = m.hexdigest()
-
         try:
             with open(config_filename) as config_file:
                 config = yaml.safe_load(config_file)
+                if "run_id" in config:
+                    self.run_id = config["run_id"]
```
aflag, line 26: Why would we want to set the run_id in the config? Given that you also removed the access log check. Is the idea that we run multiple times under the same id?

data_pipeline_api/file_formats/parameter_file.py
```diff
@@ -0,0 +1,109 @@
...
+# ======================================================================================
+
+ParameterComponent = Dict[str, Any]
+Estimate = Real
+Distribution = Union[stats.rv_discrete, stats.rv_continuous]
+Samples = np.ndarray
+
+
+class Type(Enum):
+    point_estimate = "point-estimate"
```
aflag: Aren't those usually all capitals to indicate a constant?

data_pipeline_api/file_formats/parameter_file.py
```diff
@@ -0,0 +1,109 @@
...
+def write_estimate(file: TextIOBase, component: str, estimate: Estimate):
+    write_parameter(file, component, {"type": "point-estimate", "value": estimate})
+
+
+# ======================================================================================
+# Distribution
+# ======================================================================================
+
+distribution_parsers = {
+    "gamma": lambda data: stats.gamma(a=data["shape"], scale=data["scale"]),
```
aflag, line 66: Not sure if relevant, but the distributions we are actually using are:

- Poisson
- Multinomial and binomial
- Hypergoemetric

Although I don see us reading the parameters of those distributions from this file, as they are sort of dynamically created (I'm considering the number of trials and array of probabilities to be the parameters here). @FiodorG may have better ideas.

data_pipeline_api/file_formats/object_file.py
```diff
@@ -0,0 +1,55 @@
...
+
+def get_write_group(file: IOBase, component: str) -> h5py.Group:
+    return h5py.File(file, mode="a").require_group(component)
+
+
+def read_table(file: IOBase, component: str) -> Table:
+    return pd.DataFrame(get_read_group(file, component)["table"][()]).apply(
+        lambda s: s.str.decode("utf-8")
+        if pd.api.types.is_object_dtype(s.infer_objects())
+        else s
```
aflag, line 37: Why are you transforming any objects into strings here?

data_pipeline_api/file_formats/object_file.py
```diff
@@ -0,0 +1,55 @@
...
+    )
+
+
+def write_table(file: IOBase, component: str, table: Table):
+    records = table.to_records(index=False)
```
aflag, line 42: Why are you not using df.to_hdf here?

aflag (approved):

data_pipeline_api/file_api.py
mrow84: It may not be necessary, though I have used it in the past so that I can easily locate my output for local runs. I have added a note to the spec about it, which was previously missing.

data_pipeline_api/file_formats/parameter_file.py
mrow84: yep ok, I do think it looks a bit grim but that is the convention, I'll change it

- [x] make enum values all capitals

data_pipeline_api/file_formats/object_file.py
mrow84: If you know of a better test I'm all ears/eyes - the intent is to convert `bytes` into `str`.

data_pipeline_api/file_formats/object_file.py
mrow84: I don't believe it will create the correct internal structure, and so the resulting file wouldn't conform to the API.

data_pipeline_api/file_formats/parameter_file.py
mrow84: I'll leave this for now, but I think that we should add required distributions to the spec, and then add them to the implementation.

aflag: How are we going to read parameters from them though? The array of probabilities and number of trials are calculate dynamically, aren they? How could that be in a file?

mrow84: I don't think I understand the question properly - surely if they are being dynamically created we won't be loading them?

mrow84: > How are we going to read parameters from them though?

The answer to this is that the distribution objects contain the parameters, but I still don't get the second part of what you said.

data_pipeline_api/file_formats/object_file.py
mrow84: If this is not the case, or if you have other concerns about our internal hdf5 format, do let me know.

data_pipeline_api/file_formats/parameter_file.py
aflag: What will we read from the file given that all parameters we are using right now are computed dynamically? This API assumes that you have all the parameters you need in a file before you run your simulation.

data_pipeline_api/file_formats/object_file.py
aflag: But how do you know they were not originally bytes?

data_pipeline_api/file_formats/parameter_file.py
mrow84: My understanding is that the intent of the "distribution" type is to represent a complete distribution. If you only have individual parameters then you could either read those separately, as point-estimates, or you could read a complete distribution and ignore some of its parameters.

data_pipeline_api/file_formats/object_file.py
mrow84: It is an assumption of the file format.

data_pipeline_api/file_formats/object_file.py
aflag: I just wanted to understand the reason. So the problem is that `key` is actually the component and "table" is some other concept that the to_hdf function does not expose?

data_pipeline_api/file_formats/parameter_file.py
aflag: Cool. So I think, at least for now, we are not going to be making use of these in simple network sim.

aflag (approved):

data_pipeline_api/file_formats/object_file.py
mrow84: Yes, table is the structure defined in the "standard API" spec. It isn't just that table is a concept that `to_hdf` doesn't expose, but that `to_hdf` imposes some structure of its own that is orthogonal to that required by table (afaik).

mrow84: There is some information about it in the [user guide](https://pandas.pydata.org/pandas-docs/stable/user_guide/io.html#table-format).

PR #15, opened 2020-06-27
Cppbindings Array impl and test, CMake integration

qingfengxia:


mrow84:
This needs to be carefully rebased onto master, because it has been built on a branch that was explicitly only shared for information purposes, which was squashed before merge. There are, therefore, commits here that we don't want to be added to master (for example, it appears to be adding back files in `src`). I imagine something like `git rebase 6324f880217eb8e0dcac1690dc29c1d3b505a414 --onto=master` (assuming you have pulled the latest master) should work.

qingfengxia:
My mistake, this should be sent to Ian/s branch.

bindings/cpp/README.md
```diff
@@ -54,15 +60,45 @@ This should output
...
+### Unix Makefile
 Now build the C++ test program:
 
 ```
 cd bindings/cpp
 make
 ```
+## CMake
+
+A CMakeLists.txt has been provided, it can detect system installation of pybind11, or download the latest pybind11 as a submodule.
```
ianhinder, line 76: I thought I had solved all these problems in the Makefile. The Makefile uses the current "python3" executable on the PATH to find the path to pybind11. It is tested on Ubuntu 18.04 via the CI workflow. The libraries you need to link with are also detected in the Makefile using python3-config.  The only issue I'm aware of is the problem with Python 3.8 where you *need* to specify --embed to python3-config, and in older pythons, you *can't* specify --embed (i.e. it is totally backwards-incompatible, as usual for Python).  When this came up, my plan would have been to detect explicitly Python >= 3.8 and add the --embed flag.

bindings/cpp/array.h
```diff
@@ -0,0 +1,536 @@
+// copyright, Alex Meakins and Qingfeng Xia  @ UKAEA
```
ianhinder, line 1: This seems like overkill for what we need. I also think it shouldn't be coupled to NumPy.  My original suggestion was for a short Array<T> class which essentially just does the flattening/unflattening and stores the data. You can then populate it from numpy from outside the class.

bindings/cpp/test_array.cpp
```diff
@@ -0,0 +1,115 @@
...
+#include "pybind11/numpy.h"
+#include "pybind11/stl.h"
+namespace py = pybind11;
+using namespace pybind11::literals;
+
+const std::string TEST_HDF5_FILENAME = "test_npy.h5";
+const char *TEST_DATASET_NAME = "nparray";
+
+/// std api applies here, but python open(hdf5 file is not impl, use h5py to open file
+void write_array(const string &data_product, const string &component, const Array &da)
```
ianhinder, line 16: What is the purpose of this? It should just call the python api's write_array function.   It looks like write_array has been reimplemented here, but using h5py.  There's no point implementing this here; just wait until we have the standard API writing HDF5 files, then wrap that.

bindings/cpp/test_array.cpp
```diff
@@ -0,0 +1,115 @@
...
+    attrs.attr("__setitem__")(py::str(dn + "_unit"), py::cast(dims[i].unit));
+    py::str dtitle = py::cast(dims[i].title);
+    attrs.attr("__setitem__")(py::str(dn + "_title"), dtitle);
+  }
+  // no need to close dataset?
+  h5file.attr("close")();
+}
+
+template <typename DT>
+typename ArrayT<DT>::Ptr read_array(const string &data_product, const string &component)
```
ianhinder: Same comment as write_array; this work is not needed.

bindings/cpp/test_array.cpp
mrow84: The PR implementing the HDF5 writing from python has been merged, fyi.

bindings/cpp/ParameterTypes.h
```diff
@@ -0,0 +1,73 @@
...
+
+    /// these class can be further subclass by specific model
+    class Estimation : public Parameter
+    {
+    public:
+        // ctor() set type="Estimation"
+        DT value; // T any type supported by json, can be std::vector<ET>
+    };
+
+    class Distribution : public Parameter
```
ianhinder, line 56: This probably isn't generic enough.  I think distributions will have an arbitrary number of parameters with different names, not just scale and shape.  I would probably make a Distribution out of a name, and a map<string, double>, where the latter contains the parameters of the distribution.

bindings/cpp/ParameterTypes.h
```diff
@@ -0,0 +1,73 @@
...
+    {
+    public:
+        // ctor() set type="Distribution"
+        std::string distribution;
+        double scale;
+        double shape;
+    };
+
+    /// Sample classes, plural is not a good class name
```
ianhinder, line 65: Is it necessary to introduce a new class here?  Isn't this just a vector<DT> that can be returned from read_samples?

bindings/cpp/README.md
qingfengxia: this info  deserves to make into Readme, Unix makefile section?

bindings/cpp/array.h
qingfengxia: it is also open source code, compatible with this repo,  and it is well tested. it support all numpy dtype

bindings/cpp/test_array.cpp
qingfengxia: will be remove once, rebase Chris latest code

bindings/cpp/ParameterTypes.h
qingfengxia: good idea

PR #16, opened 2020-06-30
Scrc 531 copy config yaml to access yaml

mrow84:
Address three open tickets:
[SCRC-503](https://github.com/ScottishCovidResponse/SCRCIssueTracking/issues/503)
[SCRC-505](https://github.com/ScottishCovidResponse/SCRCIssueTracking/issues/505)
[SCRC-531](https://github.com/ScottishCovidResponse/SCRCIssueTracking/issues/531)

Also fixes a bug in writing estimates (numpy floats were serialised as strings by the toml serialiser), and expands the example script.

### DIFF
--- a/data_pipeline_api/file_api.py
+++ b/data_pipeline_api/file_api.py
@@ -21,30 +21,37 @@ class FileAPI:
 
         try:
             with open(config_filename) as config_file:
-                config = yaml.safe_load(config_file)
-                if "run_id" in config:
-                    self.run_id = config["run_id"]
+                self._config = yaml.safe_load(config_file)
+                if "run_id" in self._config:
+                    self.run_id = self._config["run_id"]
                 else:
                     # Compute a run_id.
                     with open(config_filename, "rb") as file:
                         m = sha1(file.read())
                     m.update(str(self._open_timestamp).encode())
                     self.run_id = m.hexdigest()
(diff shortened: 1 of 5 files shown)


qingfengxia (approved):
some more test code added in this PR,     


`float(estimate)` should works for int, str, float, type,  if Estimate is a class, then float operator can be used to support this expr.

mrow84:
> some more test code added in this PR,
> 
> `float(estimate)` should works for int, str, float, type, if Estimate is a class, then float operator can be used to support this expr.

[`Estimate`](https://github.com/ScottishCovidResponse/data_pipeline_api/blob/8238733cb87d25d9253d991a8947f32065ce1a50/data_pipeline_api/file_formats/parameter_file.py#L14) is a [type alias](https://docs.python.org/3/library/typing.html#type-aliases) for [`Real`](https://docs.python.org/3/library/numbers.html#numbers.Real), which essentially matches `float` and `int`, but not `complex` or `Rational`.

The real problem here is that I end up passing in a `np.float64` in some circumstances (specifically when it is derived from a numpy array), which isn't explicitly supported by the toml serialiser. This solution does the coercion at a late stage, and the usage (calling with `np.float64`) isn't really compatible with the type signature, though of course because it is python that isn't obvious. Perhaps I should move the `float` coercion out, to make it clearer?

PR #20, opened 2020-07-03
SCRC-543 Add support for array labels.

mrow84:
This brings the python API in line with the full array spec. This necessitated adding a new `Array` class to encapsulate the additional functionality.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -8,7 +8,7 @@ dependencies:
   - matplotlib=3.1.3
   - pandas=1.0.3
   - toml=0.9.4
-  - h5py=2.9.0
+  - h5py=2.10.0
   - scipy=1.4.1
   - pyyaml=5.3.1
   - semver=2.9.0
--- a/data_pipeline_api/file_formats/object_file.py
+++ b/data_pipeline_api/file_formats/object_file.py
@@ -4,11 +4,44 @@ import pandas as pd
 import numpy as np
 from functools import reduce
 from operator import getitem
-from typing import Union, List
+from typing import Union, List, NamedTuple, Optional, Any
 
 
 Table = pd.DataFrame
-Array = np.ndarray
+
+
+class Dimension(NamedTuple):
+    title: Optional[str] = None
+    names: Optional[List[str]] = None
+    values: Optional[List[Any]] = None
+    units: Optional[str] = None
+
+    def __eq__(self, other):
+        if isinstance(other, Dimension):
+            return (
+                (self.title == other.title)
+                and (self.names == other.names)
+                and (self.values == other.values)
+                and (self.units == other.units)
+            )
+        else:
+            return False
+
+
+class Array(NamedTuple):
+    data: np.ndarray
+    dimensions: Optional[List[Dimension]] = None
+    units: Optional[str] = None
+
+    def __eq__(self, other):
+        if isinstance(other, Array):
+            return (
+                np.array_equal(self.data, other.data)
+                and (self.dimensions == other.dimensions)
+                and (self.units == other.units)
+            )
+        else:
+            return False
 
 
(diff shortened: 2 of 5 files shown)


environment.yml
```diff
@@ -8,7 +8,7 @@ dependencies:
   - matplotlib=3.1.3
   - pandas=1.0.3
   - toml=0.9.4
-  - h5py=2.9.0
+  - h5py=2.10.0
```
aflag, line 11: Is this needed or just keeping up with the latest versions?

qingfengxia (commented):
Array class is well designed. 

I heard all meta data like units should be H5::Attribute,  instead of H5::Dataset,   only Array will be a Dataset.  The standard spec seems have not specify it is Attribute, or Dataset.  But attribute is much easier to deal with.

While I do not know,   in my code, there is a quick switch,  either write to group or  group.attrs (AttributeProxy).   Currenlty, I thank I also write out as DataSet, maybe overkilled.   I am not sure how the R impl, R code let me daizy, haha

environment.yml
mrow84: Yes, it [broke the build](https://travis-ci.com/github/ScottishCovidResponse/data_pipeline_api/builds/174195416#L507) because of a missing member of `h5py` - it wasn't really obvious to me from the docs that it wouldn't be available in 2.9.0 (in fact it _looks_ like it _is_ available), but there we go.

mrow84:
> I heard all meta data like units should be H5::Attribute, instead of H5::Dataset, only Array will be a Dataset. The standard spec seems have not specify it is Attribute, or Dataset. But attribute is much easier to deal with.

@qingfengxia Where did you see the thing about metadata being `H5::Attribute`? It makes sense to me, but from my understanding of the [reference implementation in R](https://github.com/ScottishCovidResponse/SCRCdataAPI/blob/master/R/create_array.R) they were datasets (which is why I put it in the spec like that). It is very possible that I just misunderstood the R code.

edit: just confirmed that the R implementation uses datasets as well, so this is consistent.

qingfengxia:
@mrow84   good news to me, I do not need to make any change to my code.  I heard from Richael Reeve month earlier, he said the benefit of HDF5 is metadata rich,  small data can be added as attributes.                    while it may be old story now.   I do not understand R, so you must be right. 


Although I am still not quite get, why  unit must be plural,  Dimension_1_title, must captalized the letter D,  I am tweak my code. to be consistent to std spec.  hopefully data are exchangable between languages.

qingfengxia:
When I tried to save  std::vector<std::string>  to HDF5,  I had some several attempts. 
I even tried your  np.char.encode(), but it does not work in C++.  Finally, I found a solution, it should works for both C++ and py.

code is in  array.h,  encode_metadata().   decode_metadata(). 
``` 
                auto dt = h5py.attr("string_dtype")(); // encoding='utf-8'
                py::tuple s = py::cast(std::vector<size_t>({m_dims[i].names.size(), 1}));
                auto ds = attrs.attr("create_dataset")(py::str(dn + "_names"),
                                                       s, "dtype"_a = dt);
                int ind = 0;
                for (const auto &it : m_dims[i].names)
                {
                    ds.attr("__setitem__")(ind, py::str(it));
                    ind++;
                }
``` 

```
# python code should be 
dt = h5py.string_type()
ds = group.create_dataset(name, shape, dt)
# for each str in a list, fill it by ds[i] = str
```
The benefit of this impl,  data  can be viewed by hdf5view or h5dump.   I have no idea how R is impl,  

R:  " file.h5[[file.path(component, "Dimension_1_names")]] <- dimension_names[[1]]"

I seems save a string not a list of string.  in  that case it may be not a proper impl,

is that a formal requirement? We should write the same H5 for exchangeable.

rtblackwell (approved):

PR #28, opened 2020-07-09
Add support for using active FTP mode

rtblackwell:


data_pipeline_api/registry/common.py
```diff
@@ -147,7 +147,8 @@ def get_remote_filesystem_and_path(
         return LocalFileSystem(**storage_options), uri.as_posix()
     elif protocol in {"http", "https", "s3"}:
         # storage_options are parameters passed to request
-        uri = urllib.parse.urljoin(uri, path)
+        path = urllib.parse.quote(path)
+        uri = '/'.join(s.strip('/') for s in [uri, path])
```
mrow84, line 151: Why are you no longer using `urljoin`? Could you use `urljoin` and then `quote`, or vice versa?

data_pipeline_api/registry/common.py
```diff
@@ -156,7 +157,13 @@ def get_remote_filesystem_and_path(
         password = storage_options.pop("password", None) or inferred_options.get("password")
         uri = (Path(inferred_options["path"]) / Path(path)).as_posix()
         fs_class = FTPFileSystem if protocol == "ftp" else SFTPFileSystem
-        return fs_class(host=inferred_options["host"], username=username, password=password, **storage_options), uri
+        fs = fs_class(host=inferred_options["host"], username=username, password=password, **storage_options)
+        if protocol == "ftp":
+            try:
+                fs.ftp.dir()
+            except TimeoutError:
+                fs.ftp.set_pasv(False)
```
mrow84, line 165: I wonder if it is worth attempting the `dir` again after changing this?

mrow84 (approved):

data_pipeline_api/registry/common.py
rtblackwell: urljoin is opinionated in ways that were not helpful to me, when I just want to be able to assume that you can do a + b and make sure there's a slash

data_pipeline_api/registry/common.py
rtblackwell: I felt that that would be captured immediately on any use

PR #30, opened 2020-07-09
Java File API

vinopm:
File API functionality implemented in Java & Travis CI enabled.


vinopm:
@johnnonweiler

data_pipeline_api/java_api/src/main/java/uk/ramp/api/FileApi.java
```diff
@@ -0,0 +1,98 @@
...
+
+  /**
+   * Return file for reading corresponding to the given metadata. The file contents are hashed, and
+   * a record is made of the read.
+   *
+   * @return FileChannel for input
+   * @param query input query
+   */
+  public FileChannel openForRead(MetadataItem query) throws IOException {
+    var overridenQuery = overridesApplier.applyReadOverrides(query);
```
rtblackwell: trivial: overridden

data_pipeline_api/java_api/src/main/java/uk/ramp/config/ConfigReader.java
```diff
@@ -0,0 +1,32 @@
...
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import uk.ramp.file.FileDirectoryNormaliser;
+import uk.ramp.reader.Reader;
+import uk.ramp.yaml.YamlReader;
+
+class ConfigReader implements Reader<ImmutableConfig> {
+  private static final String LOCATION = "config.yaml";
```
rtblackwell: not sure if we'd expect this to be constant or an arg to the model

data_pipeline_api/java_api/src/main/java/uk/ramp/access/AccessLog.java
```diff
@@ -0,0 +1,31 @@
+package uk.ramp.access;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
```
rtblackwell: probably a stupid question, but I'm curious, what's the relationship between YAML and JSON in this API, my speculation is that the yaml segments are being split into json like things that the json deserializer can then make into objects?

data_pipeline_api/java_api/src/main/java/uk/ramp/access/AccessLoggerImpl.java
```diff
@@ -0,0 +1,66 @@
...
+            .build();
+    accessEntries.add(accessEntry);
+  }
+
+  @Override
+  public void logWrite(MetadataItem callMetadata, MetadataItem writeMetadata) {
+    var accessEntry =
+        ImmutableAccessEntry.builder()
+            .type("write")
+            .timestamp(clock.instant())
```
rtblackwell: How does this timestamp get logged out to access.yaml, as I guess instant is millis since epoch? The upload/download scripts are currently in python, so we need to make sure there's compatibility

data_pipeline_api/java_api/src/main/java/uk/ramp/api/FileApi.java
vinopm: Will update, thanks

data_pipeline_api/java_api/src/main/java/uk/ramp/config/ConfigReader.java
vinopm: Right now the FileApi class takes the directory (which contains config.yaml) as an input parameter. Happy to move this up to the API level and have the config.yaml path passed in instead of the directory. Let me know. Either way, it should be trivial to change this at a later stage or in a follow up PR if we're not sure yet

data_pipeline_api/java_api/src/main/java/uk/ramp/access/AccessLog.java
vinopm: This is a perfectly sensible question to ask. These Jackson core annotation names are a bit misleading. Even though these annotations have 'Json' their name, it will work. The Jackson ObjectMapper can be configured to parse all sorts of different formats including Yaml. In our case we are using the YAMLMapper (which is a configuration of ObjectMapper for Yaml). Given all of this, for yaml mappings you would still use the core annotations which have 'Json' in their names.

data_pipeline_api/java_api/src/main/java/uk/ramp/api/FileApi.java
```diff
@@ -0,0 +1,98 @@
...
+   * @return FileChannel for output
+   * @param query input query
+   */
+  public FileChannel openForWrite(MetadataItem query) throws IOException {
+    var overridenQuery = overridesApplier.applyWriteOverrides(query);
+    var filename = overridenQuery.filename().orElseThrow();
+    var normalisedFilename = fileDirectoryNormaliser.normalisePath(filename);
+    Files.createDirectories(Path.of(fileDirectoryNormaliser.parentPath()));
+    Files.createFile(Path.of(normalisedFilename));
+    accessLogger.logWrite(query, overridenQuery);
```
mrow84: I've only just scanned this, and I'm not very familiar with modern Java so maybe I've missed something, but does the write logging happen _before_ the write, or once the file has been closed? It should be the latter, and the hash of the resulting file should be computed. Again, apologies if I've just missed something.

data_pipeline_api/java_api/src/main/java/uk/ramp/access/AccessLoggerImpl.java
vinopm: Good question, and definitely one we should discuss. 

```
timestamp:
  nano: 134173000
  epochSecond: 1594376918
```

This is how it is currently logged out. Happy to change this to a standard format. Let me know

data_pipeline_api/java_api/src/main/java/uk/ramp/api/FileApi.java
vinopm: A very good question and well spotted. You're correct. The 'write logging' happens before the write to file actually happens. We log that there has been a write access just before returning the file handle. 

I didn't realise that this was a strict requirement. This will require some fiddling around as Java doesn't let you directly override the close() method on the file handle (FileChannel objects). 

The point you mentioned about the hash of the resulting file needing to be computed again is important. 

I'll work on getting this functionality implemented. 

Thanks

mrow84: Yeah nw, it's probably the most subtle/awkward part of how this works, which is why I went looking for it. I still have to look through the rest, but I imagine it's basically fine.

data_pipeline_api/java_api/src/main/java/uk/ramp/access/AccessLoggerImpl.java
rtblackwell: https://stackoverflow.com/a/45674593

I think if you do this and have it output the "2017-08-14T12:17:47.720Z" format then that's interpretable by the python yaml loader and I assume will also be interpretable by your code without more work than that configuration?

data_pipeline_api/java_api/src/main/java/uk/ramp/access/AccessLoggerFactory.java
```diff
@@ -0,0 +1,32 @@
...
+    if (config.accessLogDisabled()) {
+      return new NoImplAccessLogger();
+    }
+    var accessLogPath = config.normalisedAccessLogPath();
+
+    BufferedWriter underlyingWriter;
+    try {
+      underlyingWriter = Files.newBufferedWriter(Path.of(accessLogPath));
+    } catch (IOException e) {
+      throw new UncheckedIOException(new IOException(e));
```
johnnonweiler: Why do you create a new IOException here?  (I have a feeling that perhaps I once knew the answer to this question, but can't remember!)

data_pipeline_api/java_api/src/test/java/uk/ramp/metadata/HashVerificationMetadataSelectorTest.java
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  @Test
+  public void testOpenForRead() {
+    var query = ImmutableMetadataItem.builder().filename("file1").verifiedHash("hash1").build();
+
+    when(metadataSelector.find(query)).thenReturn(query);
+    when(hasher.hash("file1")).thenReturn("hash1");
+
+    var hashVerificationFileApi =
```
johnnonweiler: Curious choice of variable name.

data_pipeline_api/java_api/src/test/java/uk/ramp/metadata/HashVerificationMetadataSelectorTest.java
```diff
@@ -0,0 +1,33 @@
...
+        new HashVerificationMetadataSelector(metadataSelector, hasher, true);
+
+    assertThat(hashVerificationFileApi.find(query)).isEqualTo(query.withCalculatedHash("hash1"));
+  }
```
johnnonweiler: Is there a test somewhere with an invalid hash?

data_pipeline_api/java_api/src/main/java/uk/ramp/api/FileApi.java
```diff
@@ -0,0 +1,98 @@
...
+   *
+   * @return FileChannel for output
+   * @param query input query
+   */
+  public FileChannel openForWrite(MetadataItem query) throws IOException {
+    var overriddenQuery = overridesApplier.applyWriteOverrides(query);
+    var filename = overriddenQuery.filename().orElseThrow();
+    var normalisedFilename = fileDirectoryNormaliser.normalisePath(filename);
+    Files.createDirectories(Path.of(fileDirectoryNormaliser.parentPath()));
+    Files.createFile(Path.of(normalisedFilename));
```
johnnonweiler: Could `normalisedFilename` be a path to a file in another folder which does not exist, or is there a guarantee that all files will be written to the same folder (given by `fileDirectoryNormaliser.parentPath()`)?

data_pipeline_api/java_api/src/main/java/uk/ramp/access/AccessLoggerImpl.java
vinopm: That's a nice approach. Will update, thanks.

data_pipeline_api/java_api/src/main/java/uk/ramp/config/ConfigFactory.java
```diff
@@ -0,0 +1,30 @@
...
+public class ConfigFactory {
+  private static final String LOCATION = "config.yaml";
+
+  public Config config(
+      YamlReader yamlReader,
+      HasherFactory hasherFactory,
+      FileReader fileReader,
+      Instant openTimestamp,
+      FileDirectoryNormaliser fileDirectoryNormaliser) {
+    var config = new ConfigReader(yamlReader, fileDirectoryNormaliser).read();
```
johnnonweiler: It would make more sense to me to pass the normalised path to the ConfigReader constructor (instead of `fileDirectoryNormaliser`).  This `LOCATION` is hard coded in both ConfigFactory and ConfigReader, which seems wrong.

data_pipeline_api/java_api/src/main/java/uk/ramp/hash/HasherFactory.java
```diff
@@ -0,0 +1,18 @@
...
+    return new Sha1Hasher();
+  }
+
+  public Hasher fileHasher(FileReader fileReader) {
+    return fileName -> new Sha1Hasher().hash(fileReader.read(fileName));
+  }
+
+  public Hasher fileHasher(FileReader fileReader, Instant openTimestamp) {
+    return fileName -> new Sha1Hasher().hash(fileReader.read(fileName) + openTimestamp.toString());
+  }
```
johnnonweiler: I find these fileHasher methods a bit tricky to follow.  You're returning a lambda expression, the method's return type is Hasher, and the lambda expression is using a Hasher.  Since (as far as I can tell) only one of these fileHasher methods is ever used, and only once, I suggest removing them.

data_pipeline_api/java_api/src/main/java/uk/ramp/access/AccessLoggerImpl.java
vinopm: @rtblackwell Made the suggested change and by default it outputs in this format: "2020-07-10T14:08:10.367415Z". Would this be ok, or is it necessary for the seconds/millis be truncated to 3 digits?

rtblackwell: I think that will work fine
```
import yaml
yaml.safe_load("2020-07-10T14:08:10.367415Z")
datetime.datetime(2020, 7, 10, 14, 8, 10, 367415, tzinfo=datetime.timezone.utc)
```

data_pipeline_api/java_api/src/main/java/uk/ramp/config/ConfigFactory.java
```diff
@@ -0,0 +1,30 @@
...
+    var failOnHashMismatch = config.failOnHashMisMatch().orElse(true);
+    var freshHash =
+        hasherFactory
+            .fileHasher(fileReader, openTimestamp)
+            .hash(fileDirectoryNormaliser.normalisePath(LOCATION));
+    var runId = config.runId().orElse(freshHash);
+    return config
+        .withFailOnHashMisMatch(failOnHashMismatch)
```
johnnonweiler: Why do you set failOnHashMismatch here, when it is already set?

data_pipeline_api/java_api/src/main/java/uk/ramp/file/FileReader.java
```diff
@@ -0,0 +1,16 @@
...
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import org.apache.commons.io.IOUtils;
+
+public class FileReader {
+  public String read(String fileName) {
+    try (java.io.FileReader fileReader = new java.io.FileReader(new File(fileName))) {
+      return IOUtils.toString(fileReader);
+    } catch (IOException ex) {
+      throw new UncheckedIOException(ex);
```
johnnonweiler: Hmmm... it's interesting that you don't create a new IOException in this case.  (There seems to be something I'm missing here!)

data_pipeline_api/java_api/src/main/java/uk/ramp/metadata/HashVerificationMetadataSelector.java
```diff
@@ -0,0 +1,41 @@
...
+      return readMetaDataItem;
+    }
+
+    if (readMetaDataItem.filename().isEmpty()) {
+      throw new IllegalStateException("Filename not specified. " + readMetaDataItem);
+    }
+
+    var verifiedHash = readMetaDataItem.verifiedHash().get();
+
+    String calculatedHash = hasher.hash(readMetaDataItem.filename().get());
```
johnnonweiler: Was there a particular reason why you didn't use `.orElseThrow()` here, instead of checking `isEmpty()` then using `get()`?

data_pipeline_api/java_api/src/main/java/uk/ramp/config/Config.java
```diff
@@ -0,0 +1,71 @@
...
+
+  default String normalisedAccessLogPath() {
+    var accessLogPath =
+        accessLog().orElse("access-{run_id}.yaml").replace("{run_id}", runId().orElseThrow());
+
+    return normalisePath(parentPath().orElseThrow(), accessLogPath);
+  }
+
+  @JsonProperty("fail_on_hash_mismatch")
+  Optional<Boolean> failOnHashMisMatch();
```
johnnonweiler: With the immutable/json/yaml libraries you're using, is it possible to have a boolean with a default value of *true* here, rather than using an Optional \<Boolean\>?

data_pipeline_api/java_api/src/main/java/uk/ramp/metadata/MatchingMetadataSelector.java
```diff
@@ -0,0 +1,21 @@
...
+
+  MatchingMetadataSelector(List<MetadataItem> metadataItems) {
+    this.metadataItems = metadataItems;
+  }
+
+  @Override
+  public MetadataItem find(MetadataItem key) {
+    return metadataItems.stream()
+        .filter(i -> i.isSuperSetOf(key))
+        .max(comparing(i -> i.version().orElse("")))
```
johnnonweiler: Are you confident that this comparison is going to give the correct ordering?  It looks like it is using a string comparison where it should be using a numeric comparison.  I suggest adding version "11.0.0" to your test cases.

I think there has been some discussion about what goes into version strings on zulip, but I don't know what the conclusion was.

data_pipeline_api/java_api/src/test/java/uk/ramp/metadata/MatchingMetadataSelectorTest.java
```diff
@@ -0,0 +1,57 @@
...
+
+  private final MetadataItem item2 =
+      ImmutableMetadataItem.builder()
+          .dataProduct("test")
+          .filename("testFileName1")
+          .version("2.0.0")
+          .build();
+
+  @Before
+  public void setUp() {
```
johnnonweiler: I think this setUp() could be deleted.

data_pipeline_api/java_api/src/main/java/uk/ramp/metadata/MetadataItem.java
```diff
@@ -0,0 +1,108 @@
...
+
+  @JsonProperty("verified_hash")
+  Optional<String> verifiedHash();
+
+  Optional<String> calculatedHash();
+
+  Optional<String> source();
+
+  default boolean isSuperSetOf(MetadataItem key) {
+    return key.filename().map(k -> k.equals(filename().orElse(""))).orElse(true)
```
johnnonweiler: You could a method here, not sure what to call it, to do:
```
boolean keyIsEitherNotPresentOrEqual(Optional<String> key, Optional<String> other) {
    return key.map(k -> k.equals(superset.orElse(""))).orElse(true);
}
```

data_pipeline_api/java_api/src/main/java/uk/ramp/metadata/MetadataItem.java
```diff
@@ -0,0 +1,108 @@
...
+        && key.source().map(k -> k.equals(source().orElse(""))).orElse(true);
+  }
+
+  default MetadataItem applyOverrides(List<OverrideItem> readOverrides) {
+    List<MetadataItem> overridesToApply =
+        readOverrides.stream()
+            .filter(
+                readOverrideItem ->
+                    this.isSuperSetOf(
+                        readOverrideItem.where().orElse(ImmutableMetadataItem.copyOf(this))))
```
johnnonweiler: Is `this.isSuperSetOf(
                        readOverrideItem.where().orElse(ImmutableMetadataItem.copyOf(this)))`
equivalent to
`readOverrideItem.where().map(k -> this.isSuperSetOf(k)).orElse(true)` ?

johnnonweiler (commented):
This looks like great stuff.  You have clearly been very busy!

I have only got half way through reading this, but have to pause for a bit now, so I'll share what I've got so far.  I have been trying to read it carefully - most of my comments or questions are very minor things, and could be ignored.

data_pipeline_api/java_api/src/main/java/uk/ramp/metadata/MetadataSelectorFactory.java
```diff
@@ -0,0 +1,20 @@
...
+  public MetadataSelector metadataSelector(
+      Hasher hasher,
+      YamlReader yamlReader,
+      boolean shouldVerifyHash,
+      FileDirectoryNormaliser fileDirectoryNormaliser) {
+    List<MetadataItem> metadataItems =
+        new ArrayList<>(new MetaDataReader(yamlReader, fileDirectoryNormaliser).read());
+    MetadataSelector baseMetadataSelector = new MatchingMetadataSelector(metadataItems);
+    return new HashVerificationMetadataSelector(baseMetadataSelector, hasher, shouldVerifyHash);
+  }
```
johnnonweiler: Should AccessLoggedMetadataSelector be used somewhere ?

data_pipeline_api/java_api/src/main/java/uk/ramp/overrides/OverridesApplier.java
```diff
@@ -0,0 +1,49 @@
...
+
+public class OverridesApplier {
+  private final Config config;
+
+  public OverridesApplier(Config config) {
+    this.config = config;
+  }
+
+  public MetadataItem applyWriteOverrides(MetadataItem query) {
+    List<OverrideItem> readOverrides = new ArrayList<>(config.writeQueryOverrides());
```
johnnonweiler: I suggest `readOverrides` is renamed - there are two words in English which are spelt 'read', but pronounced differently, and this potentially causes confusion here.
(Could `config.writeQueryOverrides()` be used directly below, instead of `readOverrides`?)

data_pipeline_api/java_api/src/main/java/uk/ramp/overrides/OverridesApplier.java
```diff
@@ -0,0 +1,49 @@
...
+            .use(
+                ImmutableMetadataItem.builder()
+                    .filename(query.filename().orElseGet(generatedFilename))
+                    .build())
+            .build();
+
+    return query
+        .applyOverrides(readOverrides)
+        .applyOverrides(List.of(runIdOverride))
+        .applyOverrides(List.of(filenameOverride));
```
johnnonweiler: This appears to be sometimes overriding the filename with its original value.  Is this the intention?  Is it deliberately overwriting filename values from the config file?

data_pipeline_api/java_api/src/main/java/uk/ramp/overrides/OverridesApplier.java
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  public MetadataItem applyReadOverrides(MetadataItem query) {
+    return query.applyOverrides(new ArrayList<>(config.readQueryOverrides()));
```
johnnonweiler: Why is `new ArrayList<>` needed here? (I'm just curious.)

data_pipeline_api/java_api/src/main/java/uk/ramp/reader/Reader.java
```diff
@@ -0,0 +1,5 @@
+package uk.ramp.reader;
+
+public interface Reader<T> {
```
johnnonweiler: What is the purpose of the `Reader<T>` and `Writer<T>` interfaces?
(Is it a kind of documentation thing?)

data_pipeline_api/java_api/src/test/java/uk/ramp/file/FileDirectoryNormaliserTest.java
```diff
@@ -0,0 +1,28 @@
...
+public class FileDirectoryNormaliserTest {
+  private String parentPath;
+
+  @Before
+  public void setUp() throws Exception {
+    parentPath = "parentPath/";
+  }
+
+  @Test
+  public void normaliseRelativePath() {
```
johnnonweiler: Are the names of these two tests the wrong way round?  This test is normalising an absolute path, and the following test is normalising a relative path.

data_pipeline_api/java_api/src/test/java/uk/ramp/metadata/HashVerificationMetadataSelectorTest.java
johnnonweiler: Also, you're not actually testing anywhere that you're using SHA-1 hashes, which could be important, but I guess it would quickly become obvious if you got that wrong, as soon as anyone starts using the API.

data_pipeline_api/java_api/src/main/java/uk/ramp/access/AccessLoggerFactory.java
vinopm: Woops, good spot, definitely a mistake! I'll update. Thanks!

data_pipeline_api/java_api/src/test/resources/config.yaml
```diff
@@ -0,0 +1,23 @@
+data_directory: data
+access_log: access-{run_id}.yaml
```
johnnonweiler: This access_log line looks odd, but I guess it doesn't make any difference to anything.

data_pipeline_api/java_api/src/test/java/uk/ramp/metadata/HashVerificationMetadataSelectorTest.java
vinopm: I must have missed out renaming this when refactoring my own code :). Thanks for spotting, I will update it.

johnnonweiler (commented):
It looks like you've updated everything while I've been reviewing it, so my comments are probably all out of date.

I'm concerned about the comparison of version numbers, but otherwise this all looks pretty good.

I believe you have made changes only within the java_api folder, but there are files in the top level of data_pipeline_api which will probably want updating (e.g. README.md).

I wonder if we should be thinking about targetting this to the Java version on the DiRAC HPC system.  (I think it is currently using some newer features of Java which might not be available there.)

data_pipeline_api/java_api/src/main/java/uk/ramp/hash/HasherFactory.java
vinopm: Agreed, I've simplified and updated. Please let me know if you think it's still a bit tricky to follow. Thanks.

data_pipeline_api/java_api/src/main/java/uk/ramp/file/FileReader.java
vinopm: See my other comment! The usage here is correct, the other one was a mistake

vinopm:
> It looks like you've updated everything while I've been reviewing it, so my comments are probably all out of date.
> 
> I'm concerned about the comparison of version numbers, but otherwise this all looks pretty good.
> 
> I believe you have made changes only within the java_api folder, but there are files in the top level of data_pipeline_api which will probably want updating (e.g. README.md).
> 
> I wonder if we should be thinking about targetting this to the Java version on the DiRAC HPC system. (I think it is currently using some newer features of Java which might not be available there.)

@johnnonweiler Haven't changed much, just updated mainly to reflect feedback from Chris. Please continue reviewing, your feedback so far has been great!

Rikiya has been able to run the Contact Tracing Model with Java 11 on DiRAC, so i suppose we could do the same with the Covid Sim Model?

data_pipeline_api/java_api/src/test/java/uk/ramp/metadata/HashVerificationMetadataSelectorTest.java
vinopm: Thanks John, I'll add tests in for both your points mentioned.

data_pipeline_api/java_api/src/main/java/uk/ramp/api/FileApi.java
vinopm: I will change this to:
```
Files.createDirectories(Path.of(normalisedFilename).getParent());
Files.createFile(Path.of(normalisedFilename));
```
to ensure that we're creating the directories based on the parent folder of the normalisedFilename (which could point to a different path than what is stored as the parent path in fileDirectoryNormaliser)

data_pipeline_api/java_api/src/main/java/uk/ramp/config/ConfigFactory.java
vinopm: Good spot, I agree. I will update this. Thanks John

data_pipeline_api/java_api/src/main/java/uk/ramp/metadata/HashVerificationMetadataSelector.java
vinopm: No particular reason, updated it now to use `.orElseThrow()` - saves a few lines of code now

data_pipeline_api/java_api/src/main/java/uk/ramp/config/ConfigFactory.java
vinopm: The `failOnHashMismatch` is optional and can be left out in the config file. So a few lines above I am loading the optional value for `failOnHashMismatch` and defaulting it to false and then inserting it back into the config, to ensure it is not empty after the config is loaded

data_pipeline_api/java_api/src/main/java/uk/ramp/config/Config.java
vinopm: Yes, I will change this to:

```
  @JsonProperty("fail_on_hash_mismatch")
  Optional<Boolean> internalFailOnHashMisMatch();

  @Derived
  default boolean failOnHashMisMatch() {
    return internalFailOnHashMisMatch().orElse(true);
  }
```

data_pipeline_api/java_api/src/main/java/uk/ramp/config/ConfigFactory.java
vinopm: I've removed the setting of failOnHashMismatch here now. Please see your comment about 

'With the immutable/json/yaml libraries you're using, is it possible to have a boolean with a default value of true here, rather than using an Optional <Boolean>?' 

for more info.

data_pipeline_api/java_api/src/main/java/uk/ramp/metadata/MatchingMetadataSelector.java
vinopm: Good point. I initially thought the Python implementation was doing this naively too, but just double checked and it's using a library to do the comparison. I've updated this now to make sure we are doing a proper comparison. Please review again. Thanks

data_pipeline_api/java_api/src/test/java/uk/ramp/metadata/MatchingMetadataSelectorTest.java
vinopm: Updated

data_pipeline_api/java_api/src/main/java/uk/ramp/metadata/MetadataItem.java
vinopm: Updated this, hopefully is more clear now. I generally try to avoid passing in Optionals through parameters as it is considered bad practice by some. So I have passed in empty string instead, doing something quite similar to your suggestion. Let me know what you think.

data_pipeline_api/java_api/src/main/java/uk/ramp/metadata/MetadataSelectorFactory.java
vinopm: Moved this up to the FileApi level now so that it is more visible

data_pipeline_api/java_api/src/main/java/uk/ramp/overrides/OverridesApplier.java
vinopm: Agreed, updated.

data_pipeline_api/java_api/src/main/java/uk/ramp/overrides/OverridesApplier.java
vinopm: `config.readQueryOverrides()` has the type `ImmutableOverrideItem`. The `applyOverrides()` method takes in type `List<OverrideItem>` instead of `List<ImmutableOverrideItem>`. `OverrideItem` is an interface of `ImmutableOverrideItem`, so I've done this so that the types match up.

data_pipeline_api/java_api/src/test/java/uk/ramp/file/FileDirectoryNormaliserTest.java
vinopm: You're correct. I've updated now. Thanks.

data_pipeline_api/java_api/src/test/java/uk/ramp/hash/Sha1HasherTest.java
```diff
@@ -0,0 +1,18 @@
...
+
+  @Test
+  public void testSha1Hash() {
+    var input = "test";
+
+    // Generated via http://www.sha1-online.com/
+    var expectedOutput = "a94a8fe5ccb19ba61c4c0873d391e987982fbbd3";
+
+    assertThat(new Sha1Hasher().hash(input)).isEqualTo(expectedOutput);
+  }
```
vinopm: @johnnonweiler sha1 test is here now

data_pipeline_api/java_api/src/main/java/uk/ramp/reader/Reader.java
vinopm: Removed these two interfaces, they weren't necessary. Thanks.

data_pipeline_api/java_api/src/test/resources/config.yaml
vinopm: @johnnonweiler Copied this from the original python implementation. Seems that in the python implementation they replace `{run_id}` with the actual run id. @rtblackwell @mrow84 is that correct?

mrow84: Yes, there is a description of the mechanism in the [file API spec](https://teams.microsoft.com/l/file/8B530C1E-25FE-4896-BD0C-8FC8089C5AC7?tenantId=6e725c29-763a-4f50-81f2-2e254f0133c8&fileType=docx&objectUrl=https%3A%2F%2Fgla.sharepoint.com%2Fsites%2FScottishCOVID-19ResearchConsortium%2FShared%20Documents%2FGeneral%2FCollaboration%2Fdata%20pipeline%20API%2F1.%20File%20API.docx&baseUrl=https%3A%2F%2Fgla.sharepoint.com%2Fsites%2FScottishCOVID-19ResearchConsortium&serviceName=teams&threadId=19:283c5ea4551344249cc76dd0ecaa4f74@thread.tacv2&groupId=669b45e2-a9a6-4060-ba8a-9ebd3713e367) (link may or may not work, file is in teams at: General > Collaboration > data pipeline API).

data_pipeline_api/java_api/src/main/java/uk/ramp/overrides/OverridesApplier.java
vinopm: We are overriding the filename with a generated file name (based on the data product, run id, and extension provided in the query) if the filename is not provided in the query. If the file name is provided, we just override the filename with the same filename (which has no effect). The config file is not used here to determine the filename override. 

Maybe I am missing the point you were making, could you clarify please if that's the case. Thanks!

data_pipeline_api/java_api/src/main/java/uk/ramp/metadata/MetadataItem.java
vinopm: Yes, it is. Much nicer, updated. Thanks.

vinopm (commented):
Updated PR. Please could you review again @johnnonweiler @rtblackwell @mrow84. Thanks

data_pipeline_api/java_api/src/test/resources/config.yaml
johnnonweiler: Might there be any other instances of {run_id} in the yaml file that should be replaced, or is this just for the access_log?

data_pipeline_api/java_api/src/main/java/uk/ramp/file/CleanableFileChannel.java
```diff
@@ -0,0 +1,150 @@
...
+  }
+
+  @Override
+  public long read(ByteBuffer[] dsts, int offset, int length) throws IOException {
+    return fileChannelWrapper.fileChannel.read(dsts, offset, length);
+  }
+
+  @Override
+  public long read(ByteBuffer[] dsts) throws IOException {
+    return 0;
```
johnnonweiler: This looks a bit different to all the other overridden methods.  Why is this not doing a read?

johnnonweiler (approved):
Looks good to me.  Thanks for the thorough job addressing all of my previous comments.

johnnonweiler:
I'm happy with Java 11, if Rikiya has got it working on DiRAC.  (And would be interested to know how he did that.)

data_pipeline_api/java_api/src/test/resources/config.yaml
vinopm: I believe it's an edge case specific to the access_log. The current python implementation only carries out this replacement with the access_log. @mrow84 can you confirm if this is true? Thanks

mrow84: The underlying feature is to associate output files with the run that produced them. There are two categories of output file: those produced by the model, handled by the file API write mechanism, and the access log. Both categories are identified using the run id - the former during the `open_for_write` process, and the latter as this transformation of the input filename. Ideally you _always_ want the run id in the access log filename, but we allow for the case where that isn't so (which can be useful).

I don't think the transformation is required elsewhere.

data_pipeline_api/java_api/src/main/java/uk/ramp/file/CleanableFileChannel.java
vinopm: Thanks for spotting this mistake. I've updated it.

vinopm:
@johnnonweiler  Fixed the bug you pointed out + updated two other methods to return the instance of `CleanableFileChannel` instead of the underlying `FileChannel` in order to maintain encapsulation. Please review the small change. Thanks



> I'm happy with Java 11, if Rikiya has got it working on DiRAC. (And would be interested to know how he did that.)

Yes, he has managed to run with java 11 on DiRAC. I believe he downloaded OpenJDK 11 to his DiRAC home file store, and then pointed the `$JAVA_HOME` env variable to the downloaded JDK.

data_pipeline_api/java_api/src/test/resources/config.yaml
vinopm: Thanks Chris.

johnnonweiler (approved):
Looks good to me. 

Why did you add `if (fileChannel == null) { return null; }` ?  It appears from the FileChannel documentation that I looked at that `fileChannel` should never be null.

vinopm:
> Looks good to me.
> 
> Why did you add `if (fileChannel == null) { return null; }` ? It appears from the FileChannel documentation that I looked at that `fileChannel` should never be null.

Thanks @johnnonweiler. The documentation doesn't mention it but when I had a look at the implementation, it's returning null in some cases: http://hg.openjdk.java.net/jdk/jdk11/file/tip/src/java.base/share/classes/sun/nio/ch/FileChannelImpl.java#l424 Have a look at lines 359, 424, 433.

johnnonweiler:
So, the documentation could do with some improvement, and it's clearly a good idea to look at their implementation!

vinopm:
> So, the documentation could do with some improvement, and it's clearly a good idea to look at their implementation!

Yes exactly!

vinopm:
@johnnonweiler I've realised I needed to make a few changes after giving the data API document a thorough read. Please could you review my latest changes:

1. Change API so that we pass in the actual config file path into constructor of API instead of the folder path
2. some refactoring
3. Added JsonIgnore annotations to make sure derived fields are not written out to the access_log file
4. Previously the access_log was being written out to the data_directory folder instead of using the access_log path properly. When it was specified as a relative path, it was writing out into the data_directory folder. This is fixed now so that we reference the parent path to be the same folder that the config file is in.

Hopefully this is the last set of changes for this PR, if my changes in the last 3 commits look ok

Thanks again

data_pipeline_api/java_api/src/main/java/uk/ramp/config/ConfigReader.java
vinopm: I've changed this now so that the actual config file path is passed in as an arg to the File API

johnnonweiler (approved):
Looks good to me.  (I think there _might_ be cases where you're returning a String, and it would make more sense to return a Path, but it is good as it is.)

johnnonweiler:
Incidentally, is there a plan to update files in the top level of data_pipeline_api, such as README.md, in a future PR?

vinopm:
@johnnonweiler agreed, I'll address standardising the return types to `Path` in the next PR. I'll update the README.md in a future PR as well.

@mrow84 @rtblackwell could we have this merged? 

Thanks

mrow84:
> Incidentally, is there a plan to update files in the top level of data_pipeline_api, such as README.md, in a future PR?

At the moment there is only [this ticket](https://github.com/ScottishCovidResponse/SCRCIssueTracking/issues/415) about adding example usage and [this ticket](https://github.com/ScottishCovidResponse/SCRCIssueTracking/issues/548) documenting the usage of the tagging mechanism (which is currently python specific, I think). We should certainly add more into the readme.

mrow84:
I am slightly uncomfortable with the directory structure - I kind of feel like it might be better separated at the top level, but I am happy to do that later if you don't think it is going to conflict. In particular, I was wondering what will happen when we [package our code](https://github.com/ScottishCovidResponse/data_pipeline_api/blob/master/setup.py#L41) in python - @aflag will have a more informed opinion. You could try building the python package locally on your branch and checking what is included.

There are some new features that have been added since you started working on this, but I am in favour of getting this merged and adding those separately.

mrow84 (approved):

vinopm:
@mrow84 Moved it one level up now

vinopm:
@mrow84 Thanks for the approval. I will add the other changes such as functionality for `set_metadata` in a follow up PR. Could you go ahead and merge if you're comfortable. I don't have permissions to merge. Thanks

PR #31, opened 2020-07-10
SCRC-416 Add explanatory logging to file API.

mrow84:
Also, removed SimpleNetworkSimAPI, which has been added to the simple_network_sim repo, and added a note to say that the API does not permit users to read their previous writes.

This change entailed a bit of refactoring so that I could log the messages I want, which has made it a bit more involved than I otherwise would have liked.


mrow84:
Apologies for the commit noise. Codacy was extremely disapproving of my changes.

data_pipeline_api/file_api.py
```diff
@@ -1,91 +1,162 @@
...
-                    (override.get("where", {}), override.get("use", {}))
-                    for override in self._config.get("read", ())
+                logger.log(
+                    DEBUG if self.fail_on_hash_mismatch else WARNING,
+                    "fail_on_hash_mismatch = %s",
+                    self.fail_on_hash_mismatch,
+                )
+
+                self._read_overrides = FileAPI.construct_overrides(
+                    self._config.get("read", ())
```
rtblackwell, line 104: should these be the AccessType constants?

rtblackwell (approved):

data_pipeline_api/file_api.py
mrow84: Hmm, quite possibly, though it is arguable - the link is in some sense only a coincidence, because these are just the names of config sections and don't actually propagate through to the accesses directly. Additionally, this is now in a tower of four PRs, so I'd rather not change it 🤷 .

PR #34, opened 2020-07-14
SCRC-589 Added run metadata mechanism to file API to allow us to propagate model uri and git sha to upload script.

mrow84:
Some of the attributes in the access log (e.g. the `run_id`) could feasibly go in the `metadata` section, but I have left them where they are for now.

I took this opportunity to make it so the `StandardAPI` class is no longer a subclass of `FileAPI`, preventing us from inadvertently using any `FileAPI` methods.

I have added some (fairly trivial) notes on these changes into the file and standard API specs on teams.

@ianhinder, @vinopm, @ZedThree please account for these changes in your wrapper, reimplementation, and usage (respectively).

edit: This is targeting the logging branch until that is merged, to simplify the diff. I'm not sure how github handles that, if there is some better thing to do?

### DIFF
--- a/data_pipeline_api/file_api.py
+++ b/data_pipeline_api/file_api.py
@@ -72,6 +72,8 @@ class FileAPI:
         self._open_timestamp = datetime.now()
         logger.debug("open_timestamp = %s", self._open_timestamp)
         self._accesses = []
+        self._run_metadata = {}
+
         config_filename = Path(config_filename)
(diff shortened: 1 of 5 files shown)


mrow84:
Found and fixed an issue with the context manager usage. Updated tests accordingly.

rtblackwell (approved):

PR #37, opened 2020-07-15
SCRC-618 Fix incorrect string handling.

mrow84:
Reported by @aflag. Essentially a result of insufficiently tight specification of the file format (the details of which are still under discussion), meaning that the R library was producing hdf5 files that could not be read correctly by the python library.


mrow84:
The binary file I have added is a cut-down version of the one @aflag reported the issue on (to save space).

data_pipeline_api/file_formats/object_file.py
```diff
@@ -85,6 +85,26 @@ def write_table(file: IOBase, component: str, table: Table):
...
+    return np.char.decode(np.array(array, dtype=np.string_))
+
+
+def get_string_list(array) -> List[str]:
+    """Attempt to coerce an arbitrary numpy array into a string list."""
+    return list(get_string_array(array))
+
+
+def get_single_string(array) -> str:
+    """Attempt to coerce an arbitrary numpy array into a single string."""
```
rtblackwell, line 99: minor: it's not really arbitrary, it's quite specifically shaped

rtblackwell (approved):

data_pipeline_api/file_formats/object_file.py
mrow84: I felt that "attempt" was sufficient cover for all abuses.

PR #39, opened 2020-07-15
SCRC-606 Added script to convert an access.yaml into a config.yaml.

mrow84:
I have left this intentionally quite "manual" - I'm not really convinced that trying to do something more coupled to the rest of the code is particularly worthwhile, but I am open to other opinions.

### DIFF
--- a/scripts/convert_access_to_config.py
+++ b/scripts/convert_access_to_config.py
@@ -1,0 +1,65 @@
+#!/usr/bin/env python3
+import click
+import yaml
+
+
+@click.command(context_settings=dict(max_content_width=200))
+@click.argument(
+    "access_filename",
+    default="access_log.yaml",
+    type=click.Path(exists=True, dir_okay=False),
+)
+@click.argument(
+    "config_filename", default="config.yaml", type=click.Path(dir_okay=False)
+)
+@click.option(
+    "--use-filenames",
+    is_flag=True,
+    help=(
+        "Use filenames in the reproduction config, rather than the metadata used to "
+        "find them."
+    ),
+)
+def convert_cli(access_filename, config_filename, use_filenames):
+    """Convert an access.yaml file into a config.yaml that will reproduce the same run.
+    """
+    print(f"converting {access_filename} to {config_filename}")
+    with open(access_filename, "r") as access_file:
+        access = yaml.safe_load(access_file)
+        config = {}
+        # Copy top-level configuration.
+        for key in ("data_directory", "access_log", "run_id", "fail_on_hash_mismatch"):
+            if key in access["config"]:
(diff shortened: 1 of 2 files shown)


scripts/convert_access_to_config.py
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env python3
+import click
+import yaml
+
+
+@click.command(context_settings=dict(max_content_width=200))
+@click.argument("access_filename", default="access_log.yaml")
```
rtblackwell: maybe a little nicer, you can use `type=click.Path(exists=True))` here and get some fail fast, and help will tell people it ought to be a filename, and I guess you could use `type=click.Path(exists=False))` for config.yaml

scripts/convert_access_to_config.py
```diff
@@ -0,0 +1,52 @@
...
+@click.argument("config_filename", default="config.yaml")
+@click.option("--use-filenames", is_flag=True, help="Use filenames, not metadata.")
```
rtblackwell: this help doesn't really help me understand what --use-filenames does

setup.py
```diff
@@ -49,6 +49,7 @@ def _read_requirements():
             'data_registry_post = data_pipeline_api.registry.upload:upload_cli',
             'data_registry_create_model = data_pipeline_api.registry.create_model:create_model_cli',
             'data_registry_upload_input = data_pipeline_api.registry.upload_input:upload_input_cli',
+            'convert_access_to_config = scripts.convert_access_to_config:convert_cli',
```
rtblackwell: maybe we should agree on a common prefix for all of these things, RR has in the past called things `pipeline_x` when he talks about the various CLIs

mrow84: Yep seems sensible, should I change them all now?

scripts/convert_access_to_config.py
mrow84: I will try and be more expansive. I was being extra terse because of the line wrapping (that is no longer an issue).

setup.py
rtblackwell: sure

scripts/convert_access_to_config.py
mrow84: Done, though I left the `exists=False` off from the config filename, because I have found overwriting to be useful. Maybe I could add a flag that is needed to permit overwrite (though presumably I would then have to do the checking manually)?

rtblackwell: I hadn't actually realised exists=False behaved like that, I'd assumed it was more like, "it's ok if it doesn't exist", makes sense

rtblackwell (approved):

PR #40, opened 2020-07-15
update download/upload script for the new schema

rtblackwell:
+ template for uploading a new data product


data_pipeline_api/registry/access_upload.py
```diff
@@ -298,12 +243,82 @@ def unique_posts(posts: List[YamlDict]) -> List[YamlDict]:
     return [yaml.safe_load(t) for t in set_of_yamls.keys()]
 
 
+def _get_accessibility(config):
```
mrow84, line 245: This seems to imply accessibility is an `int`. Is there some kind of definition somewhere?

data_pipeline_api/registry/access_upload.py
```diff
@@ -298,12 +243,82 @@ def unique_posts(posts: List[YamlDict]) -> List[YamlDict]:
...
+    :return: the object reference to the code repo release
+    """
+    code_repo_obj = _create_target_data_dict(
+        DataRegistryTarget.object, {DataRegistryField.description: f"{model_name}+{model_git_sha}"}
+    )
+    code_repo_release = _create_target_data_dict(
+        DataRegistryTarget.code_repo_release,
+        {
+            DataRegistryField.name: model_name,
+            DataRegistryField.version: model_git_sha,
```
mrow84: The database version field seems to be a semver, so I'm not sure if this is correct. [This conversation](https://zulip.scrc.uk/#narrow/stream/14-Data/topic/Python.20API/near/16815) is related.

data_pipeline_api/registry/access_upload.py
```diff
@@ -298,12 +243,82 @@ def unique_posts(posts: List[YamlDict]) -> List[YamlDict]:
...
+    :param posts: List of posts to the data registry, will be modified
+    :param filename: path to the file to upload
+    :param remote_uri: URI to the root of the storage for uploading
+    :param storage_options: (key, value) pairs that are passed to the remote storage, e.g. credentials
+    :param storage_root: existing reference to the storage_root that this was uploaded to
+    :return: object reference to the uploaded file
+    """
+    filename = Path(filename)
+    path = upload_to_storage(remote_uri, storage_options, filename.parent, filename)
+    with open(filename, "rb") as file:
```
mrow84: You had mentioned using the FileAPI version of this, or abstracting that out. I don't really mind if you don't, though.

data_pipeline_api/registry/access_upload.py
```diff
@@ -328,71 +344,77 @@ def upload_model_run(
...
-    )
+    accessibility = _get_accessibility(config)
+    namespace = config.get("namespace")
     model_version_str = config_yaml["model_version"]
     model_name = config_yaml["model_name"]
-    accessibility = _create_target_data_dict(
-        DataRegistryTarget.accessibility, {DataRegistryField.name: accessibility_name}
-    )
     open_timestamp = config_yaml["open_timestamp"]
+    model_git_sha = config["metadata"]["git_sha"]
```
mrow84, line 399: I presume the code is fine, but you could use some constants for these instead of plain strings?

data_pipeline_api/registry/access_upload.py
```diff
@@ -403,45 +425,52 @@ def upload_model_run(
...
     help="(key, value) pairs that are passed to the remote storage, e.g. credentials",
 )
-@click.option("--remote-uri-override", type=str, help=f"URI to the root of the storage to post in the registry"
-                                                      f" required if the URI to use for download from the registry"
-                                                      f" is different from that used to upload the item")
-@click.option("--accessibility", type=str, default="public", help=f"accessibility of the data, defaults to public")
+@click.option(
+    "--remote-uri-override",
+    type=str,
+    help=f"URI to the root of the storage to post in the registry"
```
mrow84: Have you just gotten into the habit of making everything a format string? I got some low level warnings from pylint about this.

data_pipeline_api/registry/upload_data_product.py
```diff
@@ -0,0 +1,170 @@
...
+        template = f.read()
+
+    data_registry = data_registry or DEFAULT_DATA_REGISTRY_URL
+    remote_uri_override = remote_uri_override or remote_uri
+    storage_root_name = storage_root_name or urllib.parse.urlparse(remote_uri_override).netloc
+    storage_root = remote_uri_override
+    remote_option = {} if remote_option is None else dict(remote_option)
+    data_product_path = Path(data_product_path)
+
+    with open(data_product_path, "rb") as f:
```
mrow84: Does feel like a common hash function would be useful.

data_pipeline_api/registry/access_upload.py
rtblackwell: it is an int yes: https://data.scrc.uk/api/storage_root/

data_pipeline_api/registry/access_upload.py
rtblackwell: is this something I'll be able to read out the access yaml?

data_pipeline_api/registry/access_upload.py
rtblackwell: copy + pasting of actual format strings...

data_pipeline_api/registry/access_upload.py
rtblackwell: I guess it's the model_version that I was reading before

data_pipeline_api/registry/download.py
```diff
@@ -83,59 +84,71 @@ def _get_data_product_version_and_components(
...
-        # first data_product found that contains the requested component
-        for dp in data_products:
-            obj = get_on_end_point(dp[DataRegistryField.object], token)
-            data_product_components = [
-                cn
-                for cn in [
-                    get_on_end_point(c, token)[DataRegistryField.name] for c in obj[DataRegistryField.components]
+    grouped_data_products = defaultdict(list)
+    for data_product in data_products_list:
+        grouped_data_products[tuple([data_product[DataRegistryField.namespace], data_product[DataRegistryField.name]])].append(data_product)
```
mrow84: I don't think you need the `tuple` call, `grouped_data_products[data_product[DataRegistryField.namespace], data_product[DataRegistryField.name]].append(data_product)` should be enough (it amounts to the "standard" auto-tuple packing syntax used elsewhere in python).

data_pipeline_api/registry/download.py
```diff
@@ -83,59 +84,71 @@ def _get_data_product_version_and_components(
...
+
+    data_product_component_pairs = []
+    for _, group in grouped_data_products.items():
```
mrow84: Is there a reason not to use `.values()`?

data_pipeline_api/registry/download.py
```diff
@@ -83,59 +84,71 @@ def _get_data_product_version_and_components(
...
+            # data_products, and for each get their component urls, get the data at that url and break on the
+            # first data_product found that contains the requested component
+            for dp in group:
+                obj = get_on_end_point(dp[DataRegistryField.object], token)
+                data_product_components = [
+                    cn
+                    for cn in [
+                        get_on_end_point(c, token)[DataRegistryField.name] for c in obj[DataRegistryField.components]
+                    ]
+                    if re.match(glob.fnmatch.translate(parsed_config.component), cn)
```
mrow84: Why do you not just use `fnmatch.fnmatch`?

tests/registry/download.py
```diff
@@ -1,11 +1,18 @@
 import io
```
mrow84: I may have missed it, but should there be a test for the glob stuff?

mrow84: Does this file get collected by pytest?

data_pipeline_api/registry/download.py
rtblackwell: I'd vaguely wondered if accessing it through glob made it a bit clearer about what was going on, but yeah I can just change it to fnmatch, probably that's fine

tests/registry/download.py
rtblackwell: there is a download test the uses the code, but I aim to refactor it and test it more heavily then, as it currently is it's so io bound that it's a pain to write a meaningful test (which is why I need to refactor it)

rtblackwell: doh

data_pipeline_api/registry/download.py
rtblackwell: no, just a product of a refactor

data_pipeline_api/registry/access_upload.py
```diff
@@ -328,73 +391,79 @@ def upload_model_run(
...
+    model_git_sha = config["metadata"]["git_sha"]
+    model_uri = config["metadata"]["uri"]
+
     inputs = []
     outputs = []
     posts = []
 
-    _, storage_root = _add_storage_type_and_root(posts, remote_uri_override, data_registry_url, token)
+    storage_root = _add_storage_root(posts, remote_uri_override, accessibility, data_registry_url, token)
+    code_repo = _add_code_repo(posts, model_name, model_version_str, model_git_sha, model_uri)
```
mrow84, line 407: It's not clear to me whether we expect to be doing this. From some conversation on zulip I got the impression that this would be done after the fact.

mrow84 (approved):
There are a few questions I still have, but I'm happy to merge and resolve them later.

data_pipeline_api/registry/access_upload.py
rtblackwell: my impression from zulip was that we should be doing it! It's not a huge deal either way

mrow84: Yeah, I might have just been confused. My understanding was that the coderepo/version stuff would happen independently from the upload, but I'm really not sure. Anyway, as you suggest, it is easy to change.

PR #46, opened 2020-07-17
Amend lint rules

bobturneruk:


bobturneruk:
@mrow84 - please merge if OK.

mrow84 (approved):
I'm somewhat surprised that `bad-continuation` is no longer needed, it was required for a [conflict with black formatting](https://github.com/psf/black/issues/48) that I had run into; perhaps some refactoring I did made it no longer necessary. I am averse to peppering the code with pylint disable comments, but I guess that might have made this no longer being necessary more obvious.

bobturneruk:
@mrow84 - sorry that was a mistake.

mrow84:
> @mrow84 - sorry that was a mistake.

Ah - I merged it because it passed, but now that you've said that I remembered that it only checks stuff in the diffs 🤦. I guess we will catch it the next time we change that code 🤷.

PR #50, opened 2020-07-20
SCRC-635 Separate out distribution encoding function.

mrow84:
You should now be able to call `data_pipeline_api.file_formats.parameter_file.encode_distribution(...)` with a distribution object to get back the encoded form.

### DIFF
--- a/tests/file_formats/test_parameter_file.py
+++ b/tests/file_formats/test_parameter_file.py
@@ -46,17 +46,29 @@ def test_write_np_float_estimate(tmp_path):
         assert parameter_file.read_estimate(TextIOWrapper(file), "test") == estimate
 
 
+def test_encode_distribution():
+    assert parameter_file.encode_distribution(stats.norm(loc=2, scale=3)) == {
+        "type": "distribution",
+        "distribution": "norm",
+        "loc": 2,
+        "scale": 3,
+    }
+
 
 def assert_distribution_roundtrip(tmp_path, distribution):
     with open(tmp_path / "test.toml", "w+b") as file:
-        parameter_file.write_distribution(TextIOWrapper(file), f"test-{distribution.dist.name}", distribution)
+        parameter_file.write_distribution(
+            TextIOWrapper(file), f"test-{distribution.dist.name}", distribution
+        )
 
     def representation(distribution):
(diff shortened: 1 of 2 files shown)


tests/file_formats/test_parameter_file.py
```diff
@@ -46,17 +46,29 @@ def test_write_np_float_estimate(tmp_path):
...
 
 
+def test_encode_distribution():
+    assert parameter_file.encode_distribution(stats.norm(loc=2, scale=3)) == {
+        "type": "distribution",
+        "distribution": "norm",
+        "loc": 2,
+        "scale": 3,
+    }
+
```
ianhinder, line 56: Are the semantics of this (i.e. the intention once the code is generalised) that
1. "type" is always "distribution" for a distribution
2. "distribution" is the name of the distribution
3. All other keys correspond to parameters of the distribution?

mrow84: Yes, the basic structure is defined in the spec on [teams](https://teams.microsoft.com/l/file/03AF05F8-DF00-417B-BA73-B152606C0CCA?tenantId=6e725c29-763a-4f50-81f2-2e254f0133c8&fileType=docx&objectUrl=https%3A%2F%2Fgla.sharepoint.com%2Fsites%2FScottishCOVID-19ResearchConsortium%2FShared%20Documents%2FGeneral%2FCollaboration%2Fdata%20pipeline%20API%2F3.%20Standardised%20data%20type%20API.docx&baseUrl=https%3A%2F%2Fgla.sharepoint.com%2Fsites%2FScottishCOVID-19ResearchConsortium&serviceName=teams&threadId=19:283c5ea4551344249cc76dd0ecaa4f74@thread.tacv2&groupId=669b45e2-a9a6-4060-ba8a-9ebd3713e367), but the actual parameterisations are undefined. The code is "generalised" in the sense that there is [more than one distribution supported](https://github.com/ScottishCovidResponse/data_pipeline_api/blob/master/data_pipeline_api/file_formats/parameter_file.py#L67-L70), though each new distribution has to be added manually. The set of distributions that must be supported is not defined, which obviously prevents more thorough specification, but I imagine they will be added on an ad-hoc basis.

rtblackwell (approved):

PR #52, opened 2020-07-22
Refactor download script to use a Downloader object

rtblackwell:
- also add support for downloading external objects


rtblackwell:
rebased onto master

data_pipeline_api/registry/downloader.py
```diff
@@ -0,0 +1,372 @@
...
+        :param data_registry_url: base url of the data registry
+        :param token: personal access token
+        """
+        self._data_directory = Path(data_directory)
+        self._data_registry_url: str = data_registry_url or os.environ.get(DATA_REGISTRY_URL, DEFAULT_DATA_REGISTRY_URL)
+        self._token: str = token or os.environ.get(DATA_REGISTRY_ACCESS_TOKEN)
+
+        self._data_products: List[Dict[Tuple[str, str], str]] = []
+        self._external_objects: List[Dict[Tuple[str, str], str]] = []
+        self._resolved_data_products: List[Dict[Tuple[str, str], Any]] = []
```
mrow84: A triviality, but it might be nice to make a more descriptive type alias for `Dict[Tuple[str, str], Any]`, given how many times it is used.

data_pipeline_api/registry/downloader.py
```diff
@@ -0,0 +1,372 @@
...
+    def _resolve_data_products(
+        self, input_blocks: List[Dict[Tuple[str, str], Any]]
+    ) -> List[Dict[Tuple[str, str], Any]]:
+        resolved = []
+        for block in input_blocks:
+            query_data = {
+                DataRegistryField.name: block[DataRegistryTarget.data_product, DataRegistryField.name],
+                DataRegistryField.namespace: block[DataRegistryTarget.namespace, DataRegistryField.url],
+            }
+            version = block[DataRegistryTarget.data_product, DataRegistryField.version]
```
mrow84: do you want this to be a `get`? (asking because of the line below)

mrow84: there are a few similar cases in the code, so I'm going to assume the answer is no, but probably worth confirming

data_pipeline_api/registry/downloader.py
```diff
@@ -0,0 +1,372 @@
...
+            }
+            metadatas.append(metadata)
+        if metadatas:
+            yaml.safe_dump(metadatas, stream)
+
+    def _write_metadata_external_object(self, stream):
+        metadatas = []
+        for block in self._resolved_external_objects:
+            metadata = {
+                "doi_or_unique_name": block[DataRegistryTarget.external_object, DataRegistryField.doi_or_unique_name],
```
mrow84: Should this be a constant?

data_pipeline_api/registry/downloader.py
```diff
@@ -0,0 +1,372 @@
...
+                MetadataKey.accessibility: block[DataRegistryTarget.storage_root, DataRegistryField.accessibility],
+                MetadataKey.version: block[DataRegistryTarget.external_object, DataRegistryField.version],
+                MetadataKey.verified_hash: block[DataRegistryTarget.storage_location, DataRegistryField.hash],
+                MetadataKey.extension: Path(block["output_filename"]).suffix[1:],
```
mrow84: constant?

data_pipeline_api/registry/downloader.py
```diff
@@ -0,0 +1,372 @@
...
+            else:
+                logger.info(f"Data is not public, skipping download")
+
+    def resolve(self):
+        """
+        Resolves all registered data products and external objects to their expanded registry data
+        """
+        logger.info(f"Resolving {len(self._data_products)} data product references")
+        while self._data_products:
+            block = self._data_products.pop()
```
mrow84, line 328: Again really a triviality, and maybe I just didn't quite get it, but your `_resolve_*` methods seem to iterate over their input as a list, but here you seem to be packaging them up individually in a list and then passing each one through (where the output list may be larger than the input list). Could you instead do it with a chain of iterators, something like this:
```python
def f(object_in):
    yield object_in + 1
    yield object_in + 2

def g(object_in):
    yield object_in * 2

def generate(f, iterator):
    for value in iterator:
        yield from f(value)

def process(functions, iterator):
    for function in functions:
        iterator = generate(function, iterator)
    return iterator

print(list(process((f, g), [1, 2, 3])))
print(list(process((g, f), [1, 2, 3])))
```

mrow84 (approved):
I've left some comments, but none of them are blocking.

PR #58, opened 2020-07-23
Java Standard API for Parameters (no functionality for hdf5 yet)

vinopm:
I have a day off tomorrow, so thought it would be worth raising a PR for the Std API, so it can be hopefully reviewed by the time I am back on Monday.

- Includes full implementation for 'parameters' (reading and writing toml parameters) + tests

- Added some boilerplate code + tests for HDF5 (reading and writing tables and arrays) - but implementation not included in this PR

- Added setRunMetadata functionality into the FileAPI as well within this PR

- Had to amend how the filename was being loaded (in the file API) as I noticed a difference in the python/java implementation.

- I am using this statement`throw new UnsupportedOperationException()` as a placeholder for boilerplate code. This appears only for the HDF5 implementation piece & will make sure an exception is thrown if the method is referenced anywhere else.


vinopm:
@johnnonweiler

java_api/file_api/src/main/java/uk/ramp/metadata/MetadataItem.java
```diff
@@ -100,57 +108,94 @@ default MetadataItem applyOverrides(List<OverrideItem> overrides) {
...
+    Function<MetadataItem, Boolean> isFilenamePresent = meta -> meta.internalFilename().isPresent();
+    Function<MetadataItem, Boolean> isComponentPresent = meta -> meta.component().isPresent();
+    Function<MetadataItem, Boolean> isDataProductPresent = meta -> meta.dataProduct().isPresent();
+    Function<MetadataItem, Boolean> isInternalVersionPresent =
+        meta -> meta.internalVersion().isPresent();
+    Function<MetadataItem, Boolean> isExtensionPresent = meta -> meta.extension().isPresent();
+    Function<MetadataItem, Boolean> isVerifiedHashPresent = meta -> meta.verifiedHash().isPresent();
+    Function<MetadataItem, Boolean> isCalculatedHashPresent =
+        meta -> meta.calculatedHash().isPresent();
+    Function<MetadataItem, Boolean> isRunIdPresent = meta -> meta.runId().isPresent();
```
rtblackwell, line 150: @mrow84 - is run id still a thing?

java_api/file_api/src/main/java/uk/ramp/metadata/MetadataItem.java
```diff
@@ -100,57 +108,94 @@ default MetadataItem applyOverrides(List<OverrideItem> overrides) {
...
+      return false;
+    }
+    if (onlyOverrideIfEmpty) {
+      return !isBaseFieldPresent;
+    }
+    return true;
+  }
+
+  private MetadataItem applyOverride(
+      MetadataItem baseMetadata, MetadataItem metadataOverride, boolean onlyOverrideIfEmpty) {
```
rtblackwell, line 134: what's the scenario in which we don't want to apply overrides?

java_api/standard_api/src/main/java/uk/ramp/objects/NumericalArray.java
```diff
@@ -0,0 +1,9 @@
+package uk.ramp.objects;
+
+public interface NumericalArray {
+  Number[] as1DArray();
+  Number[][] as2DArray();
+  Number[][][] as3DArray();
+  Number[][][][] as4DArray();
+  Number[][][][][] as5DArray();
```
rtblackwell, line 12: classic java

java_api/standard_api/src/main/java/uk/ramp/samples/Samples.java
```diff
@@ -0,0 +1,51 @@
...
+  @Override
+  @JsonIgnore
+  default Number getSample() {
+    return sampleFrom(samples());
+  }
+
+  @Override
+  @JsonIgnore
+  default Distribution getDistribution() {
+    throw new UnsupportedOperationException("Cannot produce an distribution from a samples parameter");
```
rtblackwell: minor: `a distribution`

java_api/standard_api/src/main/java/uk/ramp/samples/Samples.java
```diff
@@ -0,0 +1,51 @@
...
+
+  @JsonIgnore
+  default Number mean() {
+    double mean = samples().stream()
+        .mapToDouble(Number::doubleValue)
+        .average()
+        .orElseThrow();
+
+    if (mean % 1 == 0) {
+      return (int) mean;
```
rtblackwell: what's the scenario where we want an int here? This feels like curious behaviour to me

java_api/file_api/src/main/java/uk/ramp/metadata/MetadataItem.java
mrow84: Yes, it is used to generate unique filenames on write.

java_api/standard_api/src/main/java/uk/ramp/api/StandardApi.java
```diff
@@ -1,5 +1,127 @@
...
+        .component(component)
+        .extension("toml")
+        .build();
+
+    try(CleanableFileChannel fileChannel = fileApi.openForWrite(query)) {
+      parameterDataWriter.write(fileChannel, component, distribution);
+    }
+  }
+
+  public Number readSample(String dataProduct, String component) {
```
mrow84: This has changed in the spec to `readSamples` - see [this issue](https://github.com/ScottishCovidResponse/SCRCIssueTracking/issues/640).

java_api/file_api/src/main/java/uk/ramp/metadata/MetadataItem.java
vinopm: The scenario is when we call open_for_write() and the filename is provided in the metadata params. In this case, we will call applyOverride with onlyOverrideIfEmpty set to True. This will result in the override for generating a fresh filename based on data_product + run_id + extension to not be used if the filename is already provided as metadata in the call to open_for_write()

java_api/standard_api/src/main/java/uk/ramp/samples/Samples.java
vinopm: Thanks, updated.

java_api/standard_api/src/main/java/uk/ramp/samples/Samples.java
vinopm: Good spot, this is not needed. If the user of this API call needs an int, they will be able to call number.intValue().
Updated PR. Thanks.

java_api/standard_api/src/main/java/uk/ramp/api/StandardApi.java
vinopm: Thanks @mrow84. Updated PR. Please review again, thanks

mrow84 (approved):
Looks broadly ok to me, but I keep on zoning out reading all of the java.

rtblackwell (approved):

PR #67, opened 2020-07-27
Upload to text file table

rtblackwell:


data_pipeline_api/registry/access_upload.py
```diff
@@ -409,10 +434,14 @@ def upload_model_run(
             )
             outputs.append(object_component)
 
-    model_config_obj = _upload_file_to_storage(posts, model_config_filename, remote_uri, storage_options, storage_root)
-    submission_script_obj = _upload_file_to_storage(
-        posts, submission_script_filename, remote_uri, storage_options, storage_root
-    )
+    if text_file_table:
```
mrow84, line 437: For the data processing stuff I am doing at the moment I *think* that I may want to be able to reference an existing non-text file for the submission script (namely a file that already exists in some database), and then only upload a new text file for the config. Thoughts?

rtblackwell: I think that ought to be handled independently of this, i.e. you either have to provide a submission script or you have to provide a reference to a submission script, in which case this would be bypassed

mrow84 (approved):

PR #70, opened 2020-07-27
Support for categorical distribution + enable codecov for Java APIs

vinopm:


vinopm:
@mrow84 @rtblackwell not really sure why the code coverage report fails on the PR, seems to be affected by whats in master?

mrow84:
I was having the same problem, though it turns out you can ignore it when merging. I don't really understand what the issue is. Did you glean any answers @bobturneruk?

rtblackwell (approved):

java_api/standard_api/src/main/java/uk/ramp/distribution/Distribution.java
```diff
@@ -87,6 +95,33 @@ private RealDistribution underlyingDistribution() {
       dist.load(
           empiricalSamples().orElseThrow().stream().mapToDouble(Number::doubleValue).toArray());
       return dist;
+    } else if (internalType().equals(DistributionType.categorical)) {
+      List<Integer> bins = bins().orElseThrow();
```
mrow84: I've lost where we were discussing this, and can't remember if I mentioned this to you already, but I did wonder whether the categories should be strings, and then parsed on the client side only when deemed appropriate?

vinopm: I've added a comment [here](https://github.com/ScottishCovidResponse/SCRCIssueTracking/issues/671) where we were discussing this. I agree with you. It will require quite a bit of work to support the string range format, i'll add that as a TODO, but let's keep this format for now as a first step. Would that be ok?

mrow84: Ah yeah, that's where it was! I mean yes, if it is a lot of work then that's fine - would adding in support for plain strings also be time consuming? I mean so that the format is this:
```
[population-ages]
type = "distribution"
distribution = "categorical"
bins = ["0", "15", "25", "55", "65", "90"]
weights = [0.1759, 0.1171, 0.4029, 0.1222, 0.1819]
```
Then we can think about the ranges later. If this is more than trivial then feel free to merge.

vinopm: Changing these to strings is straightforward. However, from doing so, we would represent 0 as its own category, whereas what we really want is 0-15 to be its own category. So changing these to strings (i.e. "15" ) without the complete change to support string ranges may be confusing from an API perspective?

mrow84: Up to you - I'm just a bit worried about (effective) API lock-in, but I don't really think it's a big deal, this isn't something that is being broadcast to the world.

vinopm: @mrow84 updated PR to include implementation for your suggestion. Please could you review again. Thanks!

mrow84 (approved):

PR #71, opened 2020-07-27
SCRC-705 Added functions to get information about the git repository

mrow84:
mostly copied from simple_network_sim

### DIFF
--- a/data_pipeline_api/git_info.py
+++ b/data_pipeline_api/git_info.py
@@ -1,0 +1,56 @@
+from typing import NamedTuple, Optional
+from pathlib import Path
+try:
+    from git import Repo, InvalidGitRepositoryError  # type: ignore
+    git_installed = True
+except ImportError:
+    Repo = None
+    InvalidGitRepositoryError = None
+    git_installed = False
+
+
+class RepoInfo(NamedTuple):
+    """
+    The info needed by the data pipeline API
+    """
+
+    git_sha: str
+    uri: str
+    is_dirty: bool
+
+
+def get_repo(path: Optional[Path] = None) -> Repo:
+    """Attempt to get a repo from path, or the working directory.
+    """
+    if Repo is None:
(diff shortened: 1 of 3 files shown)


tests/test_git_info.py
```diff
@@ -0,0 +1,23 @@
...
+    get_repo_info,
+    get_path_relative_to_repo,
+    Repo,
+    RepoInfo,
+    git_installed
+)
+
+
+# We can only test this module if git is installed.
+if git_installed:
```
mrow84: I'm actually going to change this again, to check that the interface does the expected thing when git is not installed.

mrow84: done

data_pipeline_api/git_info.py
```diff
@@ -0,0 +1,56 @@
...
+        return RepoInfo(
+            git_sha=repo.head.commit.hexsha,
+            uri=next(repo.remote("origin").urls),
+            is_dirty=repo.is_dirty(),
+        )
+    except (ImportError, InvalidGitRepositoryError):
+        return RepoInfo(git_sha="", uri=default_repo, is_dirty=True)
+
+
+def get_path_relative_to_repo(path: Path) -> Path:
```
aflag, line 50: Why do you need this function?

aflag (approved):

data_pipeline_api/git_info.py
mrow84: When we write a "code run" to the database we have to include a "submission script". At the moment this will be a text file that we upload every time, but I kind of feel that for these submission scripts (and maybe for other things) we can identify a particular file within a repo, and then use that.

In that context, this function would be used to find the location of `__main__.__file__`, which, in context would be assumed to be the "submission script", and then we would associate a reference to an existing record in the database, rather than uploading a new file every time and referencing that.

PR #73, opened 2020-07-28
SCRC-642 Add run metadata section to config and simplify config loading.

mrow84:


data_pipeline_api/standard_api.py
```diff
@@ -39,19 +39,17 @@ class StandardAPI:
...
-        git_sha: str,
+        cls, config_filename: Union[Path, str], uri: str, git_sha: str,
     ):
         return cls(FileAPI(config_filename), uri, git_sha)
 
     def __init__(self, file_api: FileAPI, uri: str, git_sha: str):
         self.file_api = file_api
-        self.file_api.set_metadata("uri", uri)
-        self.file_api.set_metadata("git_sha", git_sha)
+        self.file_api.set_run_metadata("uri", uri)
```
mrow84, line 51: In a future change I will attempt to automatically discover the git uri and sha, and at that point I think I will drop this from the API - you would then be able to set these metadata keys in the code if you thought that was necessary, or leave the auto-discovery to do it if not.

rtblackwell (approved):

data_pipeline_api/standard_api.py
ianhinder: How would auto-discovery work?  In BEEPmbp, the git version (and dirty status) is collected via "git describe --dirty --always" at *compile time* and embedded in the executable. You can't guarantee that the git version of the current working directory at runtime (in case that's how you were planning to automate this) corresponds to the version used to compile the executable.  Also, on an HPC system, we would copy the executable and any dependencies to the job work directory at job submission time to ensure that it is self-contained, since the original directory might be changed by the time the job ran. We wouldn't normally also copy the .git directory there.  So we would always have to set it manually - currently it's done via the constructor.  If we have to set it manually, then I have a few questions: Is the file_api something that users should be concerned with? Should setting metadata be proxied through the standard API? I'd rather have it as-is, I think.

On a related point (maybe this is OT for this comment!), is there a way to indicate that the git version is dirty?

mrow84: That is the reason for my saying
> you would then be able to set these metadata keys in the code if you thought that was necessary, or leave the auto-discovery to do it if not

It haven't really thought it through, but I expect it would be proxied through the StandardAPI, as it is today, just not in the constructor.

Indicating the git version is dirty is achieved by [recording an issue in simple network sim](https://github.com/ScottishCovidResponse/simple_network_sim/blob/master/simple_network_sim/sampleUseOfModel.py#L46), but there is no standard.

PR #75, opened 2020-07-28
Underlying distribution should return deterministic samples

vinopm:
The underlying distribution returned by the std api should produce deterministic samples. This is important for reproducibility.


vinopm:
Closing as not required for now

PR #80, opened 2020-07-29
Scrc 642 add database file api

mrow84:


data_pipeline_api/registry/access_upload.py
```diff
@@ -68,7 +69,38 @@ def _get_input_url(
     }
     object_component = get_data(query_data, DataRegistryTarget.object_component, data_registry_url, token)
     url = object_component["url"]
-    logger.info(f"Retrieved {url} for {namespace}/{data_product_name}/{component_name}/{version}")
+    logger.info(f"Retrieved {url} for {namespace}/{data_product_name}/{version}/{component_name}")
+    return url
+
+
+def _get_external_object_url(
+    external_object_identifier: str, version: str, component_name: str, data_registry_url: str, token: str
```
rtblackwell: I'd be inclined to call this `doi_or_unique_name` rather than `external_object_identifier` for clarity

data_pipeline_api/registry/access_upload.py
```diff
@@ -463,65 +508,22 @@ def upload_model_run(
 @click.option(
     "--submission-script", required=True, type=click.Path(exists=True), help="Path to the submission script file.",
 )
```
rtblackwell, line 620: I'm not sure about removing all the remote options and forcing it to all be through env variables, combine them with args taking precedence?

rtblackwell: + token - is there any reason to remove these args?

mrow84: no I did actually mean to put them back

data_pipeline_api/registry/access_upload.py
mrow84: done

data_pipeline_api/registry/access_upload.py
mrow84: restored to as they were, which I think is actually what we want?

data_pipeline_api/registry/access_upload.py
```diff
@@ -371,29 +403,31 @@ def upload_model_run(
...
-    model_uri = config["metadata"]["uri"]
+    run_id = run_metadata[RunMetadata.run_id]
+    namespace = run_metadata.get(RunMetadata.default_output_namespace)
+    model_version_str = run_metadata[RunMetadata.model_version]
+    model_name = run_metadata[RunMetadata.model_name]
+    open_timestamp = run_metadata[RunMetadata.open_timestamp]
+    model_git_sha = run_metadata[RunMetadata.git_sha]
+    model_uri = run_metadata[RunMetadata.git_repo]
+    remote_uri = run_metadata[RunMetadata.remote_uri]
+    remote_uri_override = run_metadata[RunMetadata.remote_uri_override]
```
rtblackwell: I don't think this should be a requirement, it should fall back to remote_uri

data_pipeline_api/registry/access_upload.py
```diff
@@ -371,29 +403,31 @@ def upload_model_run(
...
+    data_registry_url = run_metadata[RunMetadata.data_registry_url]
```
rtblackwell: and neither should this? use the get_data_registry_url function if not specified?

rtblackwell (approved):

PR #83, opened 2020-07-29
SCRC-634 Prevent tests from writing to committed data products, and disable HDF timestamp recording.

mrow84:


ianhinder (approved):
Interesting mix of features in one PR! 🙂  Looks good. I like the use of the tmp_path fixture (I learned something) for the tests.  I'll have to do something similar in the C++ bindings tests.

PR #84, opened 2020-07-29
SCRC-704 All writes record latest hash.

mrow84:
This now records the write accesses as they happen but doesn't record the hash, and then calculates them all (with caching to avoid calculating any more than once) at the end. In "interface" *structure* terms, i.e. the access.yaml, this shouldn't have changed anything, though of course the hashes for a given sequence of writes might now be different.

### DIFF
--- a/data_pipeline_api/file_api.py
+++ b/data_pipeline_api/file_api.py
@@ -2,8 +2,8 @@ from io import IOBase
 from uuid import uuid4
 from datetime import datetime
 from pathlib import Path
-from typing import Union, Optional, Any, Iterable, Dict
-from functools import wraps
+from typing import Union, Optional, Any, Iterable, Dict, List
+from dataclasses import dataclass
 from hashlib import sha1
 from logging import getLogger, WARNING, DEBUG
 import yaml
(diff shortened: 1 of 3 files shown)


data_pipeline_api/file_api.py
```diff
@@ -15,12 +15,64 @@
...
+            hash_cache = {}
+        access_metadata = self.access_metadata.copy()
+        if MetadataKey.calculated_hash not in access_metadata:
+            absolute_path = self.path.resolve()
+            try:
+                access_metadata[MetadataKey.calculated_hash] = hash_cache[absolute_path]
+            except KeyError:
+                access_metadata[MetadataKey.calculated_hash] = hash_cache[
+                    absolute_path
+                ] = FileAPI.calculate_hash(absolute_path)
```
rtblackwell, line 41: this is essentially doing like a lazy setdefault?

mrow84: yeah my aim is to avoid the computation unless absolutely necessary, because obviously for large files the hash calculation will be expensive. open to a better solution.

rtblackwell (approved):

PR #88, opened 2020-07-30
Prefix upload paths with the namespace and strip whitespace from URIs

rtblackwell:


tests/registry/test_common.py
```diff
@@ -354,20 +354,22 @@ def test_unique_dicts():
     ]
 
 
-@pytest.mark.parametrize(["remote_uri", "filename", "return_path", "expected_path"], [
-    ["ssh://server/srv/ftp/", "test.txt", "/srv/ftp/test.txt", "test_abc.txt"],
-    ["ssh://server/srv/ftp/", "some/path/test.txt", "/srv/ftp/some/path/test.txt", "some/path/test_abc.txt"],
-    ["http://somewebsite.com/", "some/path/test.txt", "http://somewebsite.com/some/path/test.txt", "some/path/test_abc.txt"],
-    ["ssh://server/", "some/path/test.txt", "some/path/test.txt", "some/path/test_abc.txt"],
+@pytest.mark.parametrize(["remote_uri", "filename", "return_path", "prefix", "expected_path"], [
```
mrow84, line 357: maybe I've missed it - is there a test for the whitespace stripping?

mrow84 (approved):

tests/registry/test_common.py
rtblackwell: that actually happens at the entry point not here... it felt a bit more like a user error than something to guard against in the innards

PR #91, opened 2020-07-30
Raise ValueError only if the hash mismatch failed

rtblackwell:


mrow84 (approved):
This was a consequence of back-and-forth refactoring - this made sense inside a function that returned in the `if` clause, but no longer!

PR #94, opened 2020-07-30
SCRC-671 Add support for additional distributions.

mrow84:
I have implemented the parameterisations described [here](https://github.com/ScottishCovidResponse/SCRCIssueTracking/issues/671#issuecomment-666427981), which I have copied to the standard API spec, though I find myself unable to click the links in teams (I don't have native Word), so I'm not really sure if they worked properly.

As mentioned in that comment, I understand that using these parameterisations may cause people to have to change code - if we can agree on what parameterisation we want (and it is a relatively normal one!) then I am happy to change any of them, but I would like to make sure we are consistent.

### DIFF
--- a/data_pipeline_api/file_formats/parameter_file.py
+++ b/data_pipeline_api/file_formats/parameter_file.py
@@ -1,8 +1,8 @@
-import toml
 from io import TextIOBase
 from enum import Enum
 from numbers import Real
-from typing import Union, Sequence, Dict, Any
+from typing import Union, Dict, Any, Tuple
+import toml
 import numpy as np
 from scipy import stats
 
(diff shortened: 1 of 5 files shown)


kzscisoft:
As mentioned on https://github.com/ScottishCovidResponse/SCRCIssueTracking/issues/693 we also require the Multinomial and Binomial distributions.

mrow84:
I am adding the additional binomial and multinomial distributions.

mrow84:
> I am adding the additional binomial and multinomial distributions.

done

kzscisoft (approved):
Yes these distributions match the request

PR #95, opened 2020-07-31
SCRC-724 Make CodeRepoRelease optional for upload.

mrow84:


data_pipeline_api/registry/access_upload.py
```diff
@@ -327,21 +327,21 @@ def to_github_uri(input_uri: str, git_sha: str = "master") -> str:
 
 def _add_code_repo(
     posts: List[YamlDict],
-    model_name: str,
-    model_version: str,
     model_git_sha: str,
     model_uri: str,
+    model_name: Optional[str],
+    model_version: Optional[str],
```
rtblackwell, line 331: I think you need to make another change for this re-ordering of args, and probably change the getting of the values from the config to a `.get`?

mrow84: Ah yeah, for some reason thought I'd done that.

rtblackwell (approved):

PR #96, opened 2020-07-31
Distributions returned from api to use specified random generator

vinopm:
All distributions returned from api to use specified random generator + some clean up

### DIFF
--- a/java_api/file_api/src/main/java/uk/ramp/config/Config.java
+++ b/java_api/file_api/src/main/java/uk/ramp/config/Config.java
@@ -7,6 +7,7 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import org.immutables.value.Value.Derived;
 import org.immutables.value.Value.Immutable;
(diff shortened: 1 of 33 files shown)


vinopm:
@johnnonweiler

johnnonweiler (approved):
This looks good for now.  (And I'll be very happy if we see it working today!)

I agree with the comments in the code along the lines of "this is a bit of a hack", and think it might make sense to review this in future, and look again at moving the RNG outside of the StandardAPI.

rtblackwell (approved):

vinopm:
@mrow84 Thanks for your feedback. The data registry download script seems to expect the config.yaml file to have data_directory within run_metadata? It fails with an error when data_directory is specified at the top level and when run_metadata does not contain data_directory

mrow84:
Ah yeah, you're right, the download script does take it from the run_metadata. Does the java implementation of the file API now just use the data_directory in the run_metadata section internally? If so I will change the python implementation accordingly and update the spec. I wonder whether we should move the other parameters (e.g. run_id) inside there as well? The output run_id might be different to the input one, but I think that is fine.

vinopm:
@mrow84 Yeah, i just changed the java api in this PR to pick up the data_directory from within run_metadata. The other params such as run_id are still at the root level. We could move it, not sure what would make most sense from the API perspective

mrow84:
Ok well in the first instance I will move the data_directory in the same way you have here. We can perhaps defer the rest to a later time.

vinopm:
Agreed, thanks @mrow84

PR #100, opened 2020-07-31
Make download script use top level  data directory.

mrow84:


data_pipeline_api/file_api.py
```diff
@@ -202,7 +202,10 @@ def __init__(self, config_filename: Optional[Union[Path, str]] = None):
         self._run_metadata["run_id"] = self._run_id
         self._run_metadata["data_directory"] = str(self._unnormalised_data_directory)
         self._run_metadata["open_timestamp"] = self._open_timestamp
-        self._run_metadata.update(self._config.get("run_metadata", {}))
+        run_metadata = self._config.get("run_metadata", {})
+        if any(key in run_metadata for key in FileAPI.RESERVED_RUN_METADATA_KEYS):
+            raise ValueError(f"reserved key {key} is set in run_metadata")
```
mrow84, line 207: I am now failing if anyone puts a reserved key into the config run_metadata.

data_pipeline_api/registry/download.py
```diff
@@ -117,6 +120,8 @@ def download_from_config_file(config_filename: Union[Path, str], token: str) ->
     with open(config_filename, "r") as cf:
         config = yaml.safe_load(cf)
     run_metadata = config.get("run_metadata", {})
+    # Copy the top level data_directory into the run_metadata.
+    run_metadata[RunMetadata.data_directory] = config["data_directory"]
```
mrow84, line 124: The download script now emulates the behaviour of the file API in copying the data directory over.

data_pipeline_api/registry/download.py
```diff
@@ -62,7 +62,10 @@ def _parse_read_config(
 
 
 def download_from_configs(
-    run_metadata: Dict[str, Any], read_configs: ReadConfigs, token: str, root_dir: Optional[Union[Path, str]] = None
+    run_metadata: Dict[str, Any],
```
mrow84, line 65: black

data_pipeline_api/registry/download.py
```diff
@@ -89,7 +92,7 @@ def download_from_configs(
                 config[MetadataKey.doi_or_unique_name],
                 config.get(MetadataKey.title),
                 config.get(MetadataKey.component),
-                config.get(MetadataKey.version)
+                config.get(MetadataKey.version),
```
mrow84, line 95: black

data_pipeline_api/file_api.py
vinopm: I should probably do this too

vinopm: with regards to throwing an exception

mrow84: yeah non-critical, as long as we have the same implementation of precedence for where the output run_metadata gets generated, but I do think it would be useful.

vinopm: Ok sure, will address separately then

vinopm (approved):

PR #103, opened 2020-08-01
C++ bindings and interface to Python data_pipeline_api

ianhinder:
This PR implements a C++ interface to the Python data_pipeline_api.  The C++ code is all in a new directory /bindings/cpp. The only change outside that directory is to add a GitHub Actions workflow for testing the C++ bindings. Instructions for how to build and test the bindings are in bindings/cpp/README.md.

The implementation is incomplete, but what is implemented seems to work.  The branch is up-to-date with current master (at the time of writing).

This interface will be used by the BEEPmbp and EERA C++ models.


bindings/cpp/array.cc
```diff
@@ -0,0 +1,69 @@
...
+#include "array.hh"
+#include <vector>
+#include <string>
+#include <cassert>
+#include <stdexcept>
+
+using namespace std;
+
+template<typename T>
+Array<T>::Array(int ndims_p, vector<int> dims_p) : dims(dims_p), ndims(ndims_p)
```
peter-t-fox, line 11: It seems unnecessary to have both an `ndims_p` parameter and also a `dims_p` parameter in the constructor, since presumably the first should always be the same as the length of the second.

peter-t-fox (approved):

bindings/cpp/array.cc
ianhinder: I agree.

PR #104, opened 2020-08-13
Add in unicode and human readable labels for distributions

### LINKED ISSUE #732: Interpret in unicode and alternate human readable parameter names for distributions

richardreeve:
Closes ScottishCovidResponse/SCRCIssueTracking#732

@mrow84 I've added in some unicode and human readable labels. If you have any time it would be great if you could have a look and see if this is acceptable code - it's been > 20 years since I last used python!

### DIFF
--- a/data_pipeline_api/file_formats/parameter_file.py
+++ b/data_pipeline_api/file_formats/parameter_file.py
@@ -5,6 +5,7 @@ from typing import Union, Dict, Any, Tuple
 import toml
 import numpy as np
 from scipy import stats
+from math import sqrt
 
 # ======================================================================================
 # Common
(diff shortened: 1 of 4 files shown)


data_pipeline_api/file_formats/parameter_file.py
```diff
@@ -5,6 +5,7 @@
 import toml
 import numpy as np
 from scipy import stats
+from math import sqrt
```
mrow84, line 8: numpy also has a square root (`np.sqrt`) and python also allows non-integer powers (i.e. `x ** 0.5`), but I don't see any reason to change it from what it is.

mrow84 (approved):

